PR #9, opened 2017-06-02
Fix nuspec namespace

mbp:
Apparently there are multiple variants of the namespace for the nuspec file

https://docs.microsoft.com/en-us/nuget/schema/nuspec uses `http://schemas.microsoft.com/packaging/2010/07/nuspec.xsd` as example.

This might not be a complete list, but solves it for my projects at least.

### DIFF
--- a/Project2015To2017/NugetPackageTransformation.cs
+++ b/Project2015To2017/NugetPackageTransformation.cs
@@ -32,8 +32,16 @@ namespace Project2015To2017
                     nuspec = XDocument.Load(filestream);
                 }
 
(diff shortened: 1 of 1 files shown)


hvanbakel:
Would you be able to add test cases for these?

mbp:
Hmm, well there is already a test (still passing) - obviously it only tests for one variant of the namespace.

If I should follow the same pattern of test, I would have to create a new TestFiles folder complete with csproj and nuspec file (only changing the namespace). I think it might be a bit too much... or what do you think?

hvanbakel:
Agreed that's a little overkill. Thanks for the great additions!

PR #56, opened 2018-02-15
Solution parsing added

almazsr:
I have added nuget package for solution parsing feature. Also, i changed `if` conditions in Main method to one `switch`, so now program can handle .sln files as well.

### DIFF
--- a/Project2015To2017/Program.cs
+++ b/Project2015To2017/Program.cs
@@ -1,4 +1,5 @@
-﻿using Project2015To2017.Definition;
+﻿using Onion.SolutionParser.Parser;
+using Project2015To2017.Definition;
 using Project2015To2017.Writing;
 using System;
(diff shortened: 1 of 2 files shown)


hvanbakel:
Thanks for the PR, what I don't entirely understand is how this is different from just working on a folder? I know you can have a solution with projects in multiple folders but:
1. that's a less common scenario
2. that just means you will have to run it more than once on multiple folders 

Also, the onion package you referenced doesn't cross compile into net core so that won't work.

mungojam:
Accepting solutions as an input would be useful for us. We have a few cases where there are multiple solutions in the same folder, or where some are in some folders, usually where we have broken apart a monolithic solution into some smaller ones. 

I expect to migrate our projects one solution at a time if that's going to be possible because that is the natural unit for us.

PR #57, opened 2018-03-03
Small fixes for encountered problems

pinnewip:
Avoid double assemblyinfo attributes,
use 'Version' instead of PackageVersion,
always update compile items (as compile files are include automatic, update is need to avoid double inclusion error),
remove .nuspec from project

### DIFF
--- a/Project2015To2017/Config.cs
+++ b/Project2015To2017/Config.cs
@@ -1,0 +1,7 @@
+﻿namespace Project2015To2017
+{
+    public static class Config
+    {
+        public static bool RemoveAssemblyInfo = true;
+    }
+}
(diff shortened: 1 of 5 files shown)


hvanbakel:
@pinnewip I took everything but the remove assembly config because 
1. it wasn't on the command line so the switch wasn't really a switch
2. I'm trying to not make irreversible changes when possible, yes people should use source control but I'm not the one that's going to write something that causes people to lose code

mungojam:
@hvanbakel I respect the principle of not losing code. An alternative that would work for us is to zip up all files that are edited or deleted so that then we only have to delete one file in each project to clear up if we have source control while those without still have an easy backup? It could even use a single zip file for the whole folder that is processed.

My motivation (as with my other requests) is that I'm hoping to process our 50 or so projects automatically. I'm sure I'll have some extra steps I have to add but the fewer the better.

hvanbakel:
Can you file that as a separate issue, I do think a few switches would be
worthwhile for these kind of things but the most basic usage should remain
safe.

mungojam:
Sure https://github.com/hvanbakel/CsprojToVs2017/issues/60 raised

PR #64, opened 2018-03-12
Consistent line endings

mungojam:
One file has inconsistent line endings causing warnings when opening in VS. Adding this default .gitattributes should stop it happening with future changes. Also fixing the file in question


mungojam:
I rebased it onto master as it was confusing before having the extra commit from my previous pull request. A useful exercise in git for me!

PR #69, opened 2018-03-20
Fix .csproj and directory processing regression

mungojam:
Logic error is causing it to think projects are directories and vice versa


hvanbakel:
Ugh that’s ugly. Thanks for fixing that one

PR #72, opened 2018-03-24
Preparatory work for nuget package creation

mungojam:
Added a prefix to the namespace as that will be important for anybody consuming it.

Moved the logic into a separate dll from the main console app as the DLL can then form the nuget package, or both could with the console app going into the tools folder or something.

Converted to .net standard 2.0 which then forced .net core 2 and .net 4.6.1. Not got a great reason for this, happy to revert if it's a problem, but it was just what I knew. It should target some .net standard version, but perhaps doesn't need to be 2.0.


mungojam:
That took a bit of wrangling, let me know if you'd like the commits squashed

PR #76, opened 2018-04-08
Split project reading and transforming

### LINKED ISSUE #75: Change transform processing to be sequential and immutable
I'm working on this today, it should make future changes, fixes and tests much simpler. 

Essentially it is making the `Project` the core entity which will be initially loaded from the legacy project rather than processing that in stages within each transform. The `Project` will contain the AssemblyInfo and Nuget info in the same way that it does at the moment. Each transform will take the `Project` and return a slightly transformed one which is then passed to the next transform.

Hope you are happy with it, I saw your comment on a recent fix you had to do, that it felt a bit hacky. Hopefully this change will make fixes easier and cleaner in future.

mungojam:
Also implements sequential transforms, passing result of each transform to next transform.

This makes the code easier to reason about and extend, such as supporting the addition of per-project pre and post transforms

I tried to keep refactoring to a minimum, but it was a fairly big structural change so it couldn't be avoided in places.

Fixes #75


mungojam:
Yeah I was putting a bit of my own functional style on it there and C# doesn't make it easy to do immutability in a nice way. I had assumed that if this PR got in I could then group some of the items up such that the constructors wouldn't be so huge. I was torn between doing lots of changes to make it tidy and making it as near the original as possible to make it easier to see the changes.

The benefit of using immutable in a case like this is that you are guaranteed that none of the transforms are meddling with the project that is passed in so you can easily compare the before and after in unit tests and you can safely re-run a transform without the risk of it breaking anything, since all the transforms are pure meaning that they always return the same thing given the same input. I think the Roslyn code base is very pure functional like this, but I know it isn't hugely widespread in C# (much more so in F# where immutability is the default).

You'll see I make use of builder classes in some places which meant that I didn't actually need to call the big constructors in most places. I think I was lazy and didn't use it for all the immutable classes that I made.

No rush, let me know if you'd like me to take out the immutable aspect. It wouldn't be too tricky to do, I'd change the transforms to be Actions rather than Funcs so that it was clear that they had side-effects, changing the passed in Project.

mungojam:
I've converted it back to mutable classes, hopefully that's good to go now. 

In doing so it made me realise one of the safety nets that you get with immutable which is that you can be sure that if say a transform created an altered package reference but then decided not to use it after all then there would be no risk that the original package reference would be changed. This doesn't apply to any of the current transforms but people would need to be reasonable careful with any newly created ones.

hvanbakel:
Merged, thanks for the work, nice change.

Issue #78, opened 2018-04-22
Should not try to upgrade ASP.NET projects?


conficient:
I have a solution that includes an ASP.NET project and this was converted along with the other projects in the solution when I used your most excellent tool.

I don't think ASP.NET projects should be converted just yet as I don't think they are not properly supported, see https://github.com/dotnet/project-system/issues/2670

hvanbakel:
If you have an example solution that exhibits the problem that would be
really valuable, both as a test to verify as well as to be sure what the
actual problem is

conficient:
>If you have an example solution that exhibits the problem that would be ...

@hvanbakel

Alas I can’t share the solution (private repo) but the utility should bypass standard asp.net (framework not core) projects

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

hvanbakel:
It doesn't need to be the exact solution, I assume it can be reduced to a small example?

What would be the marker that means it should NOT be converted?

conficient:
Well ASP.NET (Framework) projects identify as a library `<OutputType>Library</OutputType>` so you'll probably have to key on the _ProjectTypeGuids_ tag. Problem is different versions have different tags - there's a nice page in [CodeProject](https://www.codeproject.com/Reference/720512/List-of-Visual-Studio-Project-Type-GUIDs) that lists a lot of common type GUIDs. For example an MVC-v5 project will contain `{349c5851-65df-11da-9384-00065b846f21}`.

I had a look on the [dotnet/project-system] repo to see if they had a definitive supported/unsupported list but didn't see one. I've raised an [issue](https://github.com/dotnet/project-system/issues/3494) to see if they can add it.

I created a blank ASP.NET Framework project and copied the csproj (below)
```
<?xml version="1.0" encoding="utf-8"?>
<Project ToolsVersion="15.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <Import Project="..\packages\Microsoft.CodeDom.Providers.DotNetCompilerPlatform.1.0.8\build\net45\Microsoft.CodeDom.Providers.DotNetCompilerPlatform.props" Condition="Exists('..\packages\Microsoft.CodeDom.Providers.DotNetCompilerPlatform.1.0.8\build\net45\Microsoft.CodeDom.Providers.DotNetCompilerPlatform.props')" />
[...]

PR #79, opened 2018-04-27
Added UnsupportedProjectTypes helper and unit test. Added a test to ProjectConvert to check.

conficient:
Proposed fix for issue #78 - add check for unsupported projectTypeGuids


conficient:
I was going to include a test aspnet project file but the `ProjectConvert` does not have any unit tests.

PR #91, opened 2018-05-02
Remove props and targets that come from packages

### LINKED ISSUE #66: Remove props and targets that come from packages
Apply the same logic as for AssemblyReferences

See discussion https://github.com/hvanbakel/CsprojToVs2017/issues/61#issuecomment-374020733

mungojam:
These are imported at build time so don't need to be in the final project.

The check I added for the targets is quite conservative, looking for a specific set of properties as per the ones auto-generated by nuget.

I updated the nuget packages being used because I thought I was hitting a bug with ms test. Turns out it was actually DropBox causing my problem. Let me know if you would like me to roll back those updates.

I also pulled the nuget packages folder into the project as a property because I needed to re-use it for the new transformation.

Fixes #66

### DIFF
--- a/Project2015To2017/Definition/Project.cs
+++ b/Project2015To2017/Definition/Project.cs
@@ -1,6 +1,7 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Xml.Linq;
+using NuGet.Configuration;
 
 namespace Project2015To2017.Definition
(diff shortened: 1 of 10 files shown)


MagicAndre1981:
I've tested it and the entries are gone now

PR #93, opened 2018-05-07
Allow for files called \ObjSomething.xyz

mungojam:
Spotted this one while working on another fix.

Make sure that files starting with 'Obj' don't get picked up by the check for the obj folder.


hvanbakel:
Nice 👍

PR #95, opened 2018-05-07
Improve compile handling to reduce incorrect warnings

mungojam:
A few changes:

1. Remove the incorrect warnings when a <Compile Include=> gets converted to a <Compile Update=>
2. Improving the messages around why 'Update's are required
3. Treat resource designers the same as any other so that additional sub-elements and attributes are picked up


mungojam:
Sorry about the file format changes which make the diffs a bit useless. I think it's because I did ctrl+k+d on the file and it applied the editorconfig which presumably hadn't been applied before everywhere.

It could be worth going through all the files and auto-formatting as a one-off to apply the editorconfig settings.

PR #96, opened 2018-05-07
Add --dry-run option

### LINKED ISSUE #92: Provide a dummy-run mode
It will be useful for large conversions to have a practise run, where it doesn't actually change anything but purely goes through all the transformations, writing out all the messages. 

That way we can test compatibility and clear up any not-included-but-present files before doing the full run.

mungojam:
Using '--dry-run' as it is what apt-get uses

Fixes #92

Happy to rebase this or #95 once one or other is merged

### DIFF
--- a/Project2015To2017.Console/Program.cs
+++ b/Project2015To2017.Console/Program.cs
@@ -13,10 +13,11 @@ namespace Project2015To2017.Console
 		        return;
 	        }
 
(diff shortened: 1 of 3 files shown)


mungojam:
@hvanbakel Don't accept this, I don't think it's working

mungojam:
Spoke too soon, it's fine. I was calling it wrong

PR #98, opened 2018-05-12
Backup to folder

### LINKED ISSUE #60: Create single backup folder of all potentially affected files rather than renaming
Creating a single backup zip would make it possible to support deleting files such as packages.config and AssemblyInfo.cs and would be simpler to clear up once finished.

Alternatively make a copy of the whole folder like VS does (or did) when upgrading to a later version.

mungojam:
Fixes #60 

This is not ready to merge because it depends on the two open PRs. I'll rebase once those are in


Project2015To2017Tests/TestFiles/FileInclusion/SourceFileWithDesigner.Designer.cs
```diff
@@ -1,4 +1,14 @@
...
+//     This code was generated by a tool.
+//     Runtime Version:4.0.30319.42000
+//
+//     Changes to this file may cause incorrect behavior and will be lost if
+//     the code is regenerated.
+// </auto-generated>
+//------------------------------------------------------------------------------
+
+namespace Project2015To2017Tests.TestFiles.FileInclusion {
     using System;
```
mungojam: I'll get this reverted out and fixed to not happen again, it was put in by mistake

mungojam: decided to leave it in. I can't see an obvious way to stop it happening and it doesn't seem to do any harm

hvanbakel: Not too worried about the test files, they are as the name implies for test purposes anyway

PR #110, opened 2018-05-16
Don't stop on readonly files in converter

### LINKED ISSUE #109: Converter should no longer abort when encountering readonly file
It should throw a message and instead abort within the writer when it actually tries to write a file. This is because the user will soon be able to specify a custom check-out action.

mungojam:
Fixes #109 

I need to sort out the PRs so they are mergeable.  My first attempt fell flat! Will let you know when they are ready


mungojam:
The failing test works fine on my machine, so not sure what's up. I'll try fixing by making a copy of the file first so that there isn't a potential clash if tests are run in parallel

PR #121, opened 2018-06-09
Use .net core 2.1 instead of 1.1

mungojam:
Mainly to see if it makes appveyor work


mungojam:
No need to actually do this, though could be good to have a self-contained .net core exe so that users don't need to have .net core installed. Happy to just cancel this one as I resolved the earlier appveyor issue

hvanbakel:
I'm not going to merge this. Having this on a lower net core version will actually allow it to run in more places than with 2.1 so there's no need to do this.

PR #123, opened 2018-06-09
Better handling of optional assembly info attributes like AssemblyInformationalVersion

mungojam:


Project2015To2017/Definition/AssemblyAttributes.cs
```diff
@@ -1,16 +1,174 @@
...
-        public string InformationalVersion { get; set; }
-        public string Version { get; set; }
-        public string AssemblyName { get; set; }
-	    public string Description { get; set; }
-	    public string Configuration { get; set; }
-	    public string FileVersion { get; set; }
-    }
+	public sealed class AssemblyAttributes
+	{
+		public CompilationUnitSyntax FileContents { get; set; }
```
hvanbakel: This is not a place to use Roslyn in my opinion. We're not doing anything fancy here so this is just overhead and an additional dependency that we don't need to bring in.

hvanbakel: Actually, I see how you're using this now on the read side. We could use it there but then I would argue for loading the entire project to begin with 😄 I'll make a mental note of that as an improvement down the road. Ignore the above comment

Project2015To2017/Definition/AssemblyAttributes.cs
```diff
@@ -1,16 +1,174 @@
...
+		{
+			return fullName.Substring(0, fullName.Length - 9);
+		}
+
+		private bool Equals(AssemblyAttributes other)
+		{
+			return FileContents.ToFullString() == other.FileContents.ToFullString();
+		}
+
+		public override bool Equals(object obj)
```
hvanbakel: This is dangerous since you don't override gethashcode which you should. See [here](https://docs.microsoft.com/en-us/dotnet/csharp/misc/cs0659)

hvanbakel:
Awesome work 👍 great improvements overall. I'll publish a 1.6 version of the tool soon.

The upside of having all of this in one branch is that I only needed to merge this in order to close all your other PRs too. 

Would you mind going over the issues that you created if there are any that are already resolved now?

PR #124, opened 2018-06-18
Don't forget nuspec file location

mungojam:
A bug fix for one of my previous PRs


mungojam:
@hvanbakel I confused things a bit here by not tidying up. There are two commits, one is the needless obj filter which I should revert, the other is the actual fix that I was putting in and I think is still needed.

mungojam:
Thanks, it's now pulled in the revert, so it's ready for you to check whenever you want

PR #127, opened 2018-06-29
Add support for solutions that are not one directory up from the project.

cdonnellytx:
If a solution is specified, any projects converted will now use its actual location instead of assuming the solution is one level up.  If a folder or project is specified, the behavior is unchanged.

The company I work for has solutions stored in nonstandard locations, which was causing conversions to leave `<Reference>` tags for NuGet packages because it wasn't in the standard location.

### DIFF
--- a/Project2015To2017/Definition/Project.cs
+++ b/Project2015To2017/Definition/Project.cs
@@ -55,15 +55,30 @@ namespace Project2015To2017.Definition
 				var nuGetSettings = Settings.LoadDefaultSettings(projectFolder);
 				var repositoryPathSetting = SettingsUtility.GetRepositoryPath(nuGetSettings);
 
-				//return the explicitly set path, or if there isn't one, then assume the solution is one level
-				//above the project and therefore so is the 'packages' folder
-				var path = repositoryPathSetting ?? Path.GetFullPath(Path.Combine(projectFolder, @"..\packages"));
+				//return the explicitly set path, or if there isn't one, then use the solution's path if one was provided.
+				//Otherwise assume a solution is one level above the project and therefore so is the 'packages' folder
+				if (repositoryPathSetting != null)
+				{
+					return new DirectoryInfo(repositoryPathSetting);
+				}
+
+				if (Solution != null)
+				{
+					return Solution.NugetPackagesPath;
+				}
+
+				var path = Path.GetFullPath(Path.Combine(projectFolder, @"..\packages"));
 
 				return new DirectoryInfo(path);
 			}
 		}
 
 		public FileInfo PackagesConfigFile { get; set; }
+
+		/// <summary>
+		/// The solution in which this project was found, if any.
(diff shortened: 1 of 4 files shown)


Project2015To2017/ProjectConverter.cs
```diff
@@ -40,30 +41,27 @@ IProgress<string> progress
...
+				}
+
+				if (solution.ProjectPaths?.Count > 0)
+				{
+					foreach (var projectPath in solution.ProjectPaths)
 					{
-						if(!line.Trim().StartsWith("Project(")) {
-							continue;	
+						progress.Report("Project found: " + projectPath);
+						var fullPath = Path.Combine(solution.SolutionFolder.FullName, projectPath);
```
mungojam: Would be cool to move this path absoluting inside the SolutionReader as the fact that they are relative project paths is specific to the solution file format

Project2015To2017/Definition/Solution.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+	public sealed class Solution
+	{
+		public FileInfo FilePath { get; set; }
+		public IReadOnlyList<string> ProjectPaths { get; set; }
+		public DirectoryInfo SolutionFolder => FilePath.Directory;
+
+		/// <summary>
+		/// The directory where nuget stores its extracted packages for the solution.
+		/// In general this is the 'packages' folder within the solution oflder, but
```
mungojam: Minor typo older vs. folder

Project2015To2017/Definition/Project.cs
```diff
@@ -55,16 +55,31 @@ public DirectoryInfo NugetPackagesPath
 				var nuGetSettings = Settings.LoadDefaultSettings(projectFolder);
 				var repositoryPathSetting = SettingsUtility.GetRepositoryPath(nuGetSettings);
 
-				//return the explicitly set path, or if there isn't one, then assume the solution is one level
-				//above the project and therefore so is the 'packages' folder
-				var path = repositoryPathSetting ?? Path.GetFullPath(Path.Combine(projectFolder, @"..\packages"));
+				//return the explicitly set path, or if there isn't one, then use the solution's path if one was provided.
+				//Otherwise assume a solution is one level above the project and therefore so is the 'packages' folder
+				if (repositoryPathSetting != null)
```
mungojam: My guess is that you should actually consider solution level nuget.config first because I expect that a nuget.config in the project path or it's parents wouldn't count over the solution path if that is somewhere different.

hvanbakel: I think this change is correct, the closest path wins:
https://docs.microsoft.com/en-us/nuget/consume-packages/configuring-nuget-behavior#how-settings-are-applied

mungojam: Thanks, you're right, relative to the project not solution

PR #129, opened 2018-07-03
Keep Linked Files as an Include

### LINKED ISSUE #128: Keep Linked Files as an Include
Compile Includes with a Link are usually from an other directory, so they have to end as an Include and not as an Update.

cwuethrich:
Compile Includes with a Link are usually from an other directory, so they have to end as an Include and not as an Update.

fixes #128


cwuethrich:
I didn't want to add those two commits to this pull request. I'll try to fix this.

hvanbakel:
Nice one, thanks!

PR #130, opened 2018-07-08
Add Readme nuget badges

mungojam:


hvanbakel:
Merged, then updated to point to the correct one 😄. Good update though 👍

PR #133, opened 2018-07-09
Compile Include Bugfixes

### LINKED ISSUE #131: Remove None or EmbeddedResource ending with .cs
None or EmbeddedResource with file ending .cs should be removed from Compile list.

For example:
`<None Include="Class1.cs" />`

Should end as:
`<Compile Remove="Class1.cs" />`
`<None Include="Class1.cs" />`

### LINKED ISSUE #132: Remove SubType Code in Compile element
We have old classes with SubType Code. I don't know the reason for the SubType, but it should be ignored.

For example
`<Compile Include="Class1.cs"><SubType>Code</SubType></Compile>`
ends in 
`<Compile Update="Class1.cs"><SubType>Code</SubType></Compile>`.
This SubType should be ignored on files with ending cs.

cwuethrich:
fixes #131 
fixes #132

### DIFF
--- a/Project2015To2017/Transforms/FileTransformation.cs
+++ b/Project2015To2017/Transforms/FileTransformation.cs
@@ -31,10 +31,10 @@ namespace Project2015To2017.Transforms
 		{
 			var items = definition.IncludeItems;
 
-			var compileManualIncludes = FindNonWildcardMatchedFiles(definition.ProjectFolder, items, "*.cs", XmlNamespace + "Compile", progress);
-
 			var otherIncludes = ItemsToProject.SelectMany(x => items.Elements(XmlNamespace + x))
 											  .Where(KeepFileInclusion);
+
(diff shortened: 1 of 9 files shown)


Project2015To2017/Transforms/FileTransformation.cs
```diff
@@ -127,15 +133,30 @@ private static bool KeepFileInclusion(XElement x)
 				}
 			}
 
-			var filesInFolder = projectFolder.EnumerateFiles(wildcard, SearchOption.AllDirectories).Select(x => x.FullName.ToUpper()).ToArray();
+			var filesInFolder = projectFolder.EnumerateFiles(wildcard, SearchOption.AllDirectories).Select(x => x.FullName).ToArray();
```
hvanbakel: Why did you remove this ToUpper? That is to prevent an issue where the names of the files doesn't match what's in the csproj file. Windows doesn't care but Linux might.

hvanbakel:
Nice fixes, thanks!

Issue #126, opened 2018-06-18
Consider adding support for other project formats by using MsBuildExtras


mungojam:
I've successfully converted a couple of WPF projects by making use of https://github.com/onovotny/MSBuildSdkExtras which leaves me with a really nice tidy project file that debugs fine.

They aren't natively supported by the converter, so I had to first delete all the GUIDs to trick it and then override the SDK and delete a load of items from the project file afterwards, e.g. XAML and XAML.cs files.

I guess I could do it with the new custom transforms by removing these specific GUIDs beforehand and then overriding the SDK afterwards (probably not supported by `ProjectDefinition` at the moment but could be added.

Maybe that could be added as an option since you may not want to assume that people are happy with a third party SDK in their resultant projects. Hopefully Microsoft's own SDK will work for these projects in the next VS!

MagicAndre1981:
Does this "extra SDK" support WinForms? I noticed after converting the csproj that the property page of the WinForms solution is very limited and misses options like Remote Debug.

mungojam:
I think so, at least in the latest preview. I've asked them to clarify in the docs https://github.com/onovotny/MSBuildSdkExtras/issues/87

MagicAndre1981:
ok, I'll try this Extra thing and see how it works

MagicAndre1981:
doesn't work for WinForms. Also the[ provided Sample](https://github.com/onovotny/MSBuildSdkExtras/tree/master/TestProjects/Windows-Desktop-C%23/WinForms-Application) is not working differently , it has the same issues as the SDK from Microsoft

mungojam:
@MagicAndre1981 have you seen the mention on the readme that you have to add `<ExtrasEnableWinFormsProjectSetup>true</ExtrasEnableWinFormsProjectSetup>` for winforms? I think maybe that is just with the new version, but maybe only the new version supports WinForms. 

I've used the previous version with WPF without any extra tags.

Also worth trying restarting VS and deleting .vs and .obj folders. Sometimes that does some magic..

MagicAndre1981:
I've already added those settings and updated the csproj to use last dev build from myget, no difference.

MagicAndre1981:
I also found a similar SDK:

https://www.nuget.org/packages/MSBuild.NET.Extras.Sdk/

but here the VS2017 hangs at "Preparing Solution..." and nothing happens.

mungojam:
[...]

Issue #113, opened 2018-05-16
Debug/Release sections that are probably covered by Sdk are left in project file


mungojam:
I expect all or most of these sections that are left in come through as default from the Sdk in the project format. I haven't looked into it much yet though. I don't know if I'm going to pick this up in a PR or not. I'd like to as it would clean up the converted projects a lot.

The only thing I think we have changed from default is to switch on `<DocumentationFile>` but that could always be dealt with as a separate change.

```xml
  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
    <DebugSymbols>true</DebugSymbols>
    <DebugType>full</DebugType>
    <Optimize>false</Optimize>
    <OutputPath>bin\Debug\</OutputPath>
    <DefineConstants>DEBUG;TRACE</DefineConstants>
    <ErrorReport>prompt</ErrorReport>
    <WarningLevel>4</WarningLevel>
    <DocumentationFile>bin\Debug\OurAssembly.xml</DocumentationFile>
    <CodeAnalysisRuleSet>AllRules.ruleset</CodeAnalysisRuleSet>
    <Prefer32Bit>false</Prefer32Bit>
  </PropertyGroup>
  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
    <DebugType>pdbonly</DebugType>
    <Optimize>true</Optimize>
    <OutputPath>bin\Release\</OutputPath>
    <DefineConstants>TRACE</DefineConstants>
    <ErrorReport>prompt</ErrorReport>
    <WarningLevel>4</WarningLevel>
    <DocumentationFile>bin\Release\OurAssembly.xml</DocumentationFile>
    <CodeAnalysisRuleSet>AllRules.ruleset</CodeAnalysisRuleSet>
    <Prefer32Bit>false</Prefer32Bit>
  </PropertyGroup>
  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|x86' ">
    <DebugSymbols>true</DebugSymbols>
    <OutputPath>bin\x86\Debug\</OutputPath>
    <DefineConstants>DEBUG;TRACE</DefineConstants>
    <DebugType>full</DebugType>
    <PlatformTarget>x86</PlatformTarget>
    <CodeAnalysisUseTypeNameInSuppression>true</CodeAnalysisUseTypeNameInSuppression>
[...]

PR #136, opened 2018-07-16
Project properties simplification

andrew-boyarshin:
* Import and adapt parts of Microsoft.Build.Evaluation for precise Condition attribute parsing [MIT]
* Create ConditionEvaluator to glue Project2015To2017 and MSBuild expression trees
* Add sophisticated simplification routine to ProjectPropertiesReader to reduce duplication of defaults (handling variety of cases, including non-default project-wide property override and default conditionalpropertygroup-wide override)
* ProjectPropertiesReader: replace hacky configuration list retrieval routine with a better one using ConditionEvaluator
* Add Project.Platforms list property retrieval to match Project.Configurations
* [breaking change] Real support for multiple TargetFrameworks (Project.TargetFrameworks: IReadOnlyList<string> -> IList<string>, now always non-null, set accessor removed)
* Do not write default values of $(Configurations) and $(Platforms) in ProjectWriter
* Update tests to reflect changes (e.g. changed type of Project.TargetFrameworks)
* Create a number of tests to showcase and verify simplification behavior

### DIFF
--- a/Project2015To2017/Definition/Project.cs
+++ b/Project2015To2017/Definition/Project.cs
@@ -21,9 +21,10 @@ namespace Project2015To2017.Definition
 		public IReadOnlyList<XElement> Targets { get; set; }
 		public IReadOnlyList<XElement> BuildEvents { get; set; }
 		public IReadOnlyList<string> Configurations { get; set; }
+		public IReadOnlyList<string> Platforms { get; set; }
 		public IReadOnlyList<XElement> OtherPropertyGroups { get; set; }
 
-		public IReadOnlyList<string> TargetFrameworks { get; set; }
+		public IList<string> TargetFrameworks { get; } = new List<string>();
 		public bool AppendTargetFrameworkToOutputPath { get; set; } = true;
 		public ApplicationType Type { get; set; }
 		public bool Optimize { get; set; }
--- a/Project2015To2017/Reading/ConditionEvaluator.cs
+++ b/Project2015To2017/Reading/ConditionEvaluator.cs
@@ -1,0 +1,174 @@
+using System;
+using System.Collections.Generic;
+using System.Text.RegularExpressions;
+using Project2015To2017.Reading.Conditionals;
+
+namespace Project2015To2017.Reading
+{
+	/// <summary>
+	/// 
+	/// </summary>
+	public static class ConditionEvaluator
+	{
+		#region MSBuild Conditional routine
+
+		private static readonly Regex SinglePropertyRegex = new Regex(@"^\$\(([^\$\(\)]*)\)$", RegexOptions.Compiled);
+
+		/// <summary>
+		/// Update our table which keeps track of all the properties that are referenced
+		/// inside of a condition and the string values that they are being tested against.
+		/// So, for example, if the condition was " '$(Configuration)' == 'Debug' ", we
+		/// would get passed in leftValue="$(Configuration)" and rightValueExpanded="Debug".
+		/// This call would add the string "Debug" to the list of possible values for the
+		/// "Configuration" property.
+		///
+		/// This method also handles the case when two or more properties are being
+		/// concatenated together with a vertical bar, as in '
+		///     $(Configuration)|$(Platform)' == 'Debug|x86'
+		/// </summary>
+		public static void UpdateConditionedPropertiesTable
+		(
+			Dictionary<string, List<string>>
+				conditionedPropertiesTable, // List of possible values, keyed by property name
+			string leftValue, // The raw value on the left side of the operator
+			string rightValueExpanded // The fully expanded value on the right side
+			// of the operator.
+		)
+		{
+			if ((conditionedPropertiesTable != null) && (rightValueExpanded.Length > 0))
+			{
+				// The left side should be exactly "$(propertyname)" or "$(propertyname1)|$(propertyname2)"
+				// or "$(propertyname1)|$(propertyname2)|$(propertyname3)", etc.  Anything else,
+				// and we don't touch the table.
+
+				// Split up the leftValue into pieces based on the vertical bar character.
+				// PERF: Avoid allocations from string.Split by forming spans between 'pieceStart' and 'pieceEnd'
+				var pieceStart = 0;
+
+				// Loop through each of the pieces.
+				while (true)
+				{
+					var pieceSeparator = leftValue.IndexOf('|', pieceStart);
+					var lastPiece = pieceSeparator < 0;
+					var pieceEnd = lastPiece ? leftValue.Length : pieceSeparator;
+
+					var singlePropertyMatch =
+						SinglePropertyRegex.Match(leftValue, pieceStart, pieceEnd - pieceStart);
+
+					if (singlePropertyMatch.Success)
+					{
+						// Find the first vertical bar on the right-hand-side expression.
+						var indexOfVerticalBar = rightValueExpanded.IndexOf('|');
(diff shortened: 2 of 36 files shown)


andrew-boyarshin:
Progress for issues #113 and #126

andrew-boyarshin:
I have reimplemented ConditionEvaluator and some simplification parts.
I use simplified Microsoft.Build.Evaluation sources (MIT, yay!) to provide perfect Condition property parsing (matching real MSBuild of course). It can also provide proper condition evaluation for possible future needs. Shunting-Yard and RPN stack evaluator are no longer needed (it is far outmatched by real expression parsing).
I have also improved property simplification routine, targeting code clarity and handling additional cases.
I expect the code to be ready for review in 1 day. I am going to separate everything in multiple PRs this time to simplify reviewing.

Project2015To2017Tests/ProjectPropertiesReadTest.cs
```diff
@@ -596,5 +793,15 @@ public class ProjectPropertiesReadTest
 
 			return project;
 		}
+
+		internal static bool ValidateChildren(IEnumerable<XElement> value, params string[] expected)
```
andrew-boyarshin: Intentionally `internal`. Used in another PR's tests.

andrew-boyarshin:
@hvanbakel @mungojam All 3 PRs are ready to be reviewed and merged. They can work separately, order of merging is not fixed.

Project2015To2017/Reading/Conditionals/CharacterUtilities.cs
```diff
@@ -0,0 +1,34 @@
...
+        }
+
+        static internal bool IsSimpleStringChar(char candidate)
+        {
+            return (IsSimpleStringStart(candidate) || char.IsDigit(candidate));
+        }
+
+        static internal bool IsHexAlphabetic(char candidate)
+        {
+            return (candidate == 'a' || candidate == 'b' || candidate == 'c' || candidate == 'd' || candidate == 'e' || candidate == 'f' ||
```
hvanbakel: return (candidate >= 97 && candidate <= 102) || (candidate >= 65 && candidate <= 70)
would work too and is probably a lot faster

Project2015To2017/Definition/Project.cs
```diff
@@ -21,9 +21,10 @@ public sealed class Project
 		public IReadOnlyList<XElement> Targets { get; set; }
 		public IReadOnlyList<XElement> BuildEvents { get; set; }
 		public IReadOnlyList<string> Configurations { get; set; }
+		public IReadOnlyList<string> Platforms { get; set; }
 		public IReadOnlyList<XElement> OtherPropertyGroups { get; set; }
 
-		public IReadOnlyList<string> TargetFrameworks { get; set; }
+		public IList<string> TargetFrameworks { get; } = new List<string>();
```
hvanbakel: This doesn't need to be mutable, right? You read it once and from that point on it's immutable.

Project2015To2017/Reading/ConditionEvaluator.cs
```diff
@@ -0,0 +1,172 @@
...
+using System.Text.RegularExpressions;
+using Project2015To2017.Reading.Conditionals;
+
+namespace Project2015To2017.Reading
+{
+	public static class ConditionEvaluator
+	{
+		#region MSBuild Conditional routine
+
+		private static readonly Lazy<Regex> SinglePropertyRegex = new Lazy<Regex>(
```
hvanbakel: The cost of initializing this is not worth the Lazy<T> construct?

hvanbakel:
Looks good in general, a few comments I don't know if you want to update your PR to reflect those or if you want me to fix those when I merge it down?

I'll also be doing a few style changes (I very much dislike if blocks without braces) but that's taste.

Project2015To2017/Definition/Project.cs
andrew-boyarshin: But I do add new items there. First, it is used in ProjectReader in 2nd PR. then, it is used in ProjectPropertiesReader for TargetFrameworkVersion. I believe mutability is nothing compared to how comparatively ugly things would be  with constructing new colection, preserving old ones and adding new.

Project2015To2017/Reading/Conditionals/CharacterUtilities.cs
andrew-boyarshin: Well, that's MSBuild. :) I really don't think we should bring in our own modifications except those to reduce dependencies on MSBuild code. We might later want to update code from upstream so such changes might introduce confusion.

hvanbakel:
Now that the other one is merged you should be able to simplify https://github.com/hvanbakel/CsprojToVs2017/blob/master/Project2015To2017/Reading/ProjectReader.cs?#L104

Project2015To2017/Reading/ConditionEvaluator.cs
andrew-boyarshin: MSBuild code again (probably to make no-op actions faster). But this is indeed sth I could change. Will do.

Project2015To2017/Reading/Conditionals/CharacterUtilities.cs
hvanbakel: Interesting, can we include a link to the source somewhere here in case things get outdated?

andrew-boyarshin:
OK, I am working on this PR (expected timeframe — 4 hours). Will also rebase on latest HEAD. I will also include XamlTransformationTest since 2nd PR is merged now.

andrew-boyarshin:
@hvanbakel done! Added also upstream.md to highlight the essentials of MSBuild code import.

Project2015To2017/Reading/ProjectPropertiesReader.cs
```diff
@@ -79,20 +92,278 @@ public static void PopulateProperties(Project project, XDocument projectXml)
...
+						                          child.Value == projectName:
+						case "AssemblyName" when !hasParentCondition && ValidateEmptyConditionValue(childCondition) &&
+						                         child.Value == projectName:
+						{
+							removeQueue.Add(child);
+							break;
+						}
+
+						// Default configuration-specific paths
+						// todo: move duplicated paths from conditional groups to non-conditional one
```
andrew-boyarshin: That is a serious matter for different PR. I have a good idea how to implement this one though.

Project2015To2017/Reading/Conditionals/CharacterUtilities.cs
andrew-boyarshin: Introduced upstream.md in *Reading\* directory.

mungojam: I think the answer is no, but would https://github.com/daveaglick/Buildalyzer be of any use here to avoid copying the code in? It would be a bit too hefty a dependency probably and may not actually help. Or are git sub-modules another possibility to directly bring in the code from MS Build? I've never actually used them before.

andrew-boyarshin: @mungojam unfortunately, we cannot bring in the code directly. It requires removal of specific code, non-trivial replacements in a couple of places so that there are no dependencies introduced. And since we don't want our code to bring in the whole enormous dependency on MSBuild *and* its dependencies *and* .NET Core SDK *and* Microsoft SDKs *and* NuGet, Buildalyzer (that does exactly the thing) can be of no use to us.

PR #139, opened 2018-07-23
Minor improvements

andrew-boyarshin:
* Do not allow generated AssemblyInfo in case more that 1 AssemblyInfo file is found
* Add Project.HasMultipleAssemblyInfos flag
* Add test to ensure GenerateAssemblyInfo is set to false correctly
* Do not remove Microsoft.CSharp assembly reference (#40 is wrong, it is still needed)
* Update tests to reflect changes
* Minor formatting fixes

### DIFF
--- a/Project2015To2017/Definition/Project.cs
+++ b/Project2015To2017/Definition/Project.cs
@@ -36,6 +36,8 @@ namespace Project2015To2017.Definition
 		public FileInfo FilePath { get; set; }
 		public DirectoryInfo ProjectFolder => FilePath.Directory;
 
+		public bool HasMultipleAssemblyInfoFiles { get; set; }
+
 		/// <summary>
 		/// Files or folders that should be deleted as part of the conversion
 		/// </summary>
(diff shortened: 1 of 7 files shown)


andrew-boyarshin:
Regarding Microsoft.CSharp. I better support my statement with extract from project.assets.json (.NET Core 2.1, .NET Framework 4.7.2, MSBuild Preview 15.8 and so on).
```json
".NETCoreApp,Version=v2.1": {
  "Microsoft.CSharp/4.5.0": {
    "type": "package",
    "compile": {
      "ref/netcoreapp2.0/_._": {}
    },
    "runtime": {
      "lib/netcoreapp2.0/_._": {}
    }
  },
},
".NETFramework,Version=v4.7.2": {
  "Microsoft.CSharp/4.5.0": {
    "type": "package",
    "frameworkAssemblies": [
      "Microsoft.CSharp"
    ],
    "compile": {
      "ref/net45/_._": {}
    },
    "runtime": {
      "lib/net45/_._": {}
    }
  },
}
```
As we can see, .NET Core 2.0+ (from [sources](https://github.com/dotnet/corefx/blob/master/src/Microsoft.CSharp/pkg/Microsoft.CSharp.pkgproj)) does not require Microsoft.CSharp, .NET Framework does (uses GAC version). Without using Microsoft.CSharp reference dynamic (as language feature) code does not compile for .NET Framework (as I have experienced for myself).

hvanbakel:
Thanks for this, great work on the `Microsoft.CSharp` reference. I've found things like that hard to find.

PR #140, opened 2018-07-26
Introduce caching and minor refactorings

andrew-boyarshin:
* Cache loaded projects and condition evaluation states (controlled by switch below)
* Introduce ProjectReader.EnableCaching property that defaults to false (so no surprises in tests, but enabled in Console project)
* Fix Project2015To2017.Console.csproj OutputType
* Store data source XML elements for future changes (to be able later to reference data source precisely)
* Preparation for future PRs (more information to come)
* Nuget -> NuGet
* Use ProjectReference for Solution model (to reduce future code duplication and simplify semantics)
* Add resolved project file path to ProjectReference (to simplify use of ProjectReference in Project creation)
* Refactor solution handling in ProjectConverter.Convert (introduce ConvertSolution method)
* Use the only solution in current directory if present (to match MSBuild behavior)
* Fix indentation according to .editorconfig (tabs instead of spaces, I fixed only files with mixed indent)
* Greatly improve XamlPagesTransformation to account for MSBuild.Sdk.Extras advanced features
* Add Instance fields for SolutionReader and ProjectReader to reduce memory pressure on huge solutions (non-obligatory)

Behavior changes:
* If there is one and only solution in specified directory to convert, use that.
* XamlPagesTransformation now filters more elements

Future work (not this PR):
* Working analysis for common problems in project configuration
* Ability to read (with only FileTransformation enabled) already converted (CPS) projects (no writing of course) to support running analysis again and account for changed line numbers

### DIFF
--- a/Project2015To2017.Console/Options.cs
+++ b/Project2015To2017.Console/Options.cs
@@ -29,7 +29,7 @@ namespace Project2015To2017.Console
 			{
 				KeepAssemblyInfo = AssemblyInfo,
 				TargetFrameworks = TargetFrameworks?.ToList(),
-				AppendTargetFrameworkToOutputPath = !NoTargetFrameworkToOutputPath
+				AppendTargetFrameworkToOutputPath = !NoTargetFrameworkToOutputPath,
 			};
 	}
 }
--- a/Project2015To2017/Definition/Project.cs
+++ b/Project2015To2017/Definition/Project.cs
@@ -8,7 +8,8 @@ namespace Project2015To2017.Definition
 {
 	public sealed class Project
 	{
-		public static readonly XNamespace XmlNamespace = "http://schemas.microsoft.com/developer/msbuild/2003";
+		public static readonly XNamespace XmlLegacyNamespace = "http://schemas.microsoft.com/developer/msbuild/2003";
+		public static readonly XNamespace XmlNamespace = XmlLegacyNamespace;
 
 		public IReadOnlyList<AssemblyReference> AssemblyReferences { get; set; }
 		public IReadOnlyList<ProjectReference> ProjectReferences { get; set; }
@@ -23,6 +24,8 @@ namespace Project2015To2017.Definition
 		public IReadOnlyList<string> Configurations { get; set; }
 		public IReadOnlyList<string> Platforms { get; set; }
 		public IReadOnlyList<XElement> OtherPropertyGroups { get; set; }
+
+		public XDocument ProjectDocument { get; set; }
 
 		public IList<string> TargetFrameworks { get; } = new List<string>();
 		public bool AppendTargetFrameworkToOutputPath { get; set; } = true;
@@ -46,11 +49,11 @@ namespace Project2015To2017.Definition
 		public IReadOnlyList<FileSystemInfo> Deletions { get; set; }
 
 		/// <summary>
-		/// The directory where nuget stores its extracted packages for the project.
+		/// The directory where NuGet stores its extracted packages for the project.
 		/// In general this is the 'packages' folder within the parent solution, but
 		/// it can be overridden, which is accounted for here.
 		/// </summary>
-		public DirectoryInfo NugetPackagesPath
+		public DirectoryInfo NuGetPackagesPath
 		{
 			get
 			{
@@ -68,7 +71,7 @@ namespace Project2015To2017.Definition
 
 				if (this.Solution != null)
 				{
-					return this.Solution.NugetPackagesPath;
+					return this.Solution.NuGetPackagesPath;
 				}
 
 				var path = Path.GetFullPath(Path.Combine(projectFolder, "..", "packages"));
--- a/Project2015To2017/ProjectConverter.cs
+++ b/Project2015To2017/ProjectConverter.cs
@@ -1,11 +1,11 @@
+using Project2015To2017.Definition;
+using Project2015To2017.Reading;
+using Project2015To2017.Transforms;
 using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Runtime.CompilerServices;
-using Project2015To2017.Definition;
-using Project2015To2017.Reading;
-using Project2015To2017.Transforms;
 
 [assembly: InternalsVisibleTo("Project2015To2017Tests")]
 
@@ -31,14 +31,14 @@ namespace Project2015To2017
 			};
 		}
 
-		public static IEnumerable<Definition.Project> Convert(
+		public static IEnumerable<Project> Convert(
 			string target,
 			IProgress<string> progress)
 		{
 			return Convert(target, new List<ITransformation>(), new List<ITransformation>(), progress);
 		}
 
-		public static IEnumerable<Definition.Project> Convert(
+		public static IEnumerable<Project> Convert(
 			string target,
 			ConversionOptions conversionOptions,
 			IProgress<string> progress)
@@ -46,7 +46,7 @@ namespace Project2015To2017
 			return Convert(target, conversionOptions, new List<ITransformation>(), new List<ITransformation>(), progress);
 		}
 
-		public static IEnumerable<Definition.Project> Convert(
+		public static IEnumerable<Project> Convert(
 			string target,
 			IReadOnlyList<ITransformation> preTransforms,
 			IReadOnlyList<ITransformation> postTransforms,
@@ -56,7 +56,7 @@ namespace Project2015To2017
 			return Convert(target, new ConversionOptions(), preTransforms, postTransforms, progress);
 		}
 
-		public static IEnumerable<Definition.Project> Convert(
+		public static IEnumerable<Project> Convert(
 			string target,
 			ConversionOptions conversionOptions,
 			IReadOnlyList<ITransformation> preTransforms,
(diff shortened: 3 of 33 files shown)


Project2015To2017.Console/Options.cs
```diff
@@ -29,7 +29,7 @@ public ConversionOptions ConversionOptions
 			{
 				KeepAssemblyInfo = AssemblyInfo,
 				TargetFrameworks = TargetFrameworks?.ToList(),
-				AppendTargetFrameworkToOutputPath = !NoTargetFrameworkToOutputPath
+				AppendTargetFrameworkToOutputPath = !NoTargetFrameworkToOutputPath,
```
andrew-boyarshin: For future PR.

Project2015To2017/Definition/Project.cs
```diff
@@ -8,7 +8,8 @@ namespace Project2015To2017.Definition
 {
 	public sealed class Project
 	{
-		public static readonly XNamespace XmlNamespace = "http://schemas.microsoft.com/developer/msbuild/2003";
+		public static readonly XNamespace XmlLegacyNamespace = "http://schemas.microsoft.com/developer/msbuild/2003";
```
andrew-boyarshin: For future PR.

Project2015To2017/ProjectConverter.cs
```diff
@@ -56,83 +56,111 @@ IProgress<string> progress
...
-				yield break;
 			}
-
-			// Process only the given project file
-			yield return ProcessFile(target, null, conversionOptions, preTransforms, postTransforms, progress);
 		}
 
-		private static Definition.Project ProcessFile(
-				string filePath,
+		private static Project ProcessFile(
```
andrew-boyarshin: To preserve API contract.

Project2015To2017/Reading/ProjectReader.cs
```diff
@@ -1,28 +1,48 @@
...
+		/// <summary>
+		/// Process-lifetime-long cache of loaded projects
+		/// <remarks>
+		/// It was supposed to be of <see cref="System.Runtime.Caching.ObjectCache"/> type, but that can still remove items from cache under memory pressure,
+		/// while we still need availability guarantees.
+		/// </remarks>
+		/// </summary>
+		private readonly Dictionary<int, Project> _cache = new Dictionary<int, Project>();
+
+		public Project Read(FileInfo file, IProgress<string> progress, bool loadModern = false)
```
andrew-boyarshin: 3rd parameter for future PR (yeah, that's support for modern CPS project reading).

Project2015To2017/Reading/ProjectReader.cs
```diff
@@ -1,28 +1,48 @@
...
+			if (EnableCaching && _cache.TryGetValue(hash, out var projectDefinition))
+			{
+				return projectDefinition;
+			}
+
 			XDocument projectXml;
 			using (var stream = File.Open(filePath, FileMode.Open, FileAccess.Read, FileShare.Read))
 			{
-				projectXml = XDocument.Load(stream);
+				projectXml = XDocument.Load(stream, LoadOptions.SetLineInfo);
```
andrew-boyarshin: To be able later to find line number for XML element data source stored in new fields in a couple of classes.

Project2015To2017/Reading/ProjectReader.cs
```diff
@@ -32,39 +52,36 @@ public Project Read(string filePath, IProgress<string> progress)
 				return null;
 			}
 
-			XNamespace nsSys = "http://schemas.microsoft.com/developer/msbuild/2003";
-			if (projectXml.Element(nsSys + "Project") == null)
+			var isLegacy = projectXml.Element(XmlLegacyNamespace + "Project") != null;
```
andrew-boyarshin: For future PR.

Project2015To2017/Reading/ProjectReader.cs
```diff
@@ -32,39 +52,36 @@ public Project Read(string filePath, IProgress<string> progress)
...
 				Deletions = Array.Empty<FileSystemInfo>(),
 				AssemblyAttributeProperties = Array.Empty<XElement>()
 			};
 
+			if (EnableCaching)
+			{
+				_cache.Add(hash, projectDefinition);
+			}
+			
+			projectDefinition.AssemblyReferences = LoadAssemblyReferences(projectXml, progress);
```
andrew-boyarshin: Moved these to be able in future PR to pass `projectDefinition` as parameter instead of `projectXml`.

Project2015To2017/Transforms/FileTransformation.cs
```diff
@@ -156,15 +156,19 @@ private static bool KeepFileInclusion(XElement x)
 				knownFullPaths.Add(Path.GetFullPath(Path.Combine(projectFolder.FullName, otherIncludeMatchingWildcard)));
 			}
 
-			foreach (var nonListedFile in filesInFolder.Except(knownFullPaths, StringComparer.OrdinalIgnoreCase))
+			// if (!project.IsModernProject)
```
andrew-boyarshin: The only change there is additional indentation level for future `if` statement.

Project2015To2017/Transforms/Helpers.cs
```diff
@@ -1,22 +1,42 @@
...
-		    return source.Elements().Where(e => e.Name.LocalName == localName);
-	    }
-    }
+		public static IEnumerable<XElement> ElementsAnyNamespace<T>(this T source, string localName)
+			where T : XContainer
+		{
+			return source.Elements().Where(e => e.Name.LocalName == localName);
+		}
+		
+		public static string GetRelativePathTo(this FileSystemInfo from, FileSystemInfo to)
```
andrew-boyarshin: Not used now, used in future PR.

Project2015To2017/Writing/ProjectWriter.cs
```diff
@@ -218,6 +218,11 @@ internal XElement CreateXml(Project project)
 						projectReferenceElement.Add(new XElement("Aliases", projectReference.Aliases));
 					}
 
+					if (projectReference.EmbedInteropTypes)
```
andrew-boyarshin: Setting this property to its correct value didn't make it into this PR due to dependency on future PR. Why not add writing the value now (though never executed at the moment).

Project2015To2017Tests/XamlTransformationTest.cs
```diff
@@ -120,17 +120,17 @@ public class XamlTransformationTest
 
 			// App.xaml is NOT included due to ApplicationDefinition
 			// App.xaml.cs is NOT included due to <SubType>Code</SubType> (FileTransformation)
-			// Views\Shell.xaml.cs is included due to DependentUpon
-			// .\..\Views\Initialize.xaml.cs is included due to DependentUpon
+			// Views\Shell.xaml.cs is NOT included due to Compile+DependentUpon
```
andrew-boyarshin: Improved due to improvements in XamlPagesTransformation.

Project2015To2017/Reading/ProjectReader.cs
mungojam: I like where I think that's going. I've converted a few projects and then wished I could run improved versions of my custom transformations, or take advantage of improvements to the converter.

Project2015To2017/ProjectConverter.cs
```diff
@@ -56,83 +56,111 @@ IProgress<string> progress
...
-					}
+					yield return project;
 				}
 
 				yield break;
 			}
 
 			// Process all csprojs found in given directory
-			if (!Path.GetExtension(target).Equals(".csproj", StringComparison.OrdinalIgnoreCase))
+			var projectFiles = Directory.EnumerateFiles(target, "*.csproj", SearchOption.AllDirectories).ToArray();
```
mungojam: Given the above change to process single solutions, feels like maybe this should get caught up in that too. Let's say the user has a folder with just 10 sub-folders each with 1 solution in. I think it could be done with recursion over the previous conditions.

It messes with my head a bit to think what the most logical thing is though. What you've currently gone for feels a bit more inconsistent and unpredictable than it was before. The option is there to process a solution if that's what people want to do.

mungojam: Maybe an option could be introduced 'process-orphan-projects' or something like that

Project2015To2017/Reading/ProjectReader.cs
```diff
@@ -1,28 +1,48 @@
...
+		/// It was supposed to be of <see cref="System.Runtime.Caching.ObjectCache"/> type, but that can still remove items from cache under memory pressure,
+		/// while we still need availability guarantees.
+		/// </remarks>
+		/// </summary>
+		private readonly Dictionary<int, Project> _cache = new Dictionary<int, Project>();
+
+		public Project Read(FileInfo file, IProgress<string> progress, bool loadModern = false)
 		{
+			var filePath = file.FullName;
+			var hash = filePath.GetHashCode();
```
mungojam: You can just use the filePath as the key in the Dictionary. It will calculate the hashcode internally anyway as that is how it does fast lookups.

mungojam:
It's not my place to say for this project, but I found a good article that sounded like a good and efficient way to propose multiple changes like this. It means that maintainers can easily process each change one at a time without you having to do lots of remerging later, just branch of each PR in turn:

https://graysonkoonce.com/stacked-pull-requests-keeping-github-diffs-small/

Project2015To2017/Reading/ProjectReader.cs
andrew-boyarshin: Yeah, I know that, but I considered that usually hashtables do store full keys too to verify that there is no collision on adding a new key to the hashtables. I considered that memory efficiency is more important than collision protection and store only hashcode.

Project2015To2017/ProjectConverter.cs
andrew-boyarshin: I have always wondered why it would recursively scan for projects. I was surprised when I ran csproj-to-2017 half a year ago in a directory with a solution and a lot of subprojects not intended to be converted. I believe we should respect solution file more and blunt recursive tree descent less.

andrew-boyarshin:
@hvanbakel review, please?

Project2015To2017/ProjectConverter.cs
hvanbakel: I think the origin of that is the natural progression of:
1. pass me a project file (too much work specifying individual ones)
2. pass me a directory (recursively search projects which would work for a solution)
3. only at this point solution support was added
I still think that the directory support is useful, but that comes down to taste. It did become less useful when solution support was added.

Project2015To2017/Reading/ProjectReader.cs
hvanbakel: I would keep the Dictionary to be `Dictionary<FileInfo, Project>` that says more about what you're storing than the hashcode. I think that code readability is more important than perf here, I have seen solutions with ~150 projects but that becomes pretty unwieldy so the memory overhead is minimal here, we're not storing millions of entries.

Project2015To2017/Reading/ConditionEvaluator.cs
```diff
@@ -119,31 +119,60 @@ public static void UpdateConditionedPropertiesTable
...
+
 		public static Dictionary<string, List<string>> GetConditionValues(string condition)
 		{
-			var state = new ConditionEvaluationStateImpl();
+			if (condition.Length > 0)
+			{
+				condition = condition.Trim();
+			}
+
+			var hash = condition.GetHashCode();
```
hvanbakel: same as in the other one, I would keep this the original type that is the key, having a hashcode in there requires intimate knowlegde of how to do these comparisons while it won't add a lot of benefit

Project2015To2017/Reading/ProjectReader.cs
```diff
@@ -1,28 +1,48 @@
...
 using System.Xml.Linq;
-using Project2015To2017.Definition;
 using static Project2015To2017.Definition.Project;
 
 namespace Project2015To2017.Reading
 {
 	public class ProjectReader
 	{
-		public Project Read(string filePath)
+		public static readonly ProjectReader Instance = new ProjectReader();
```
hvanbakel: I think we should leave this an instance but then initialize it with either a `Project` or a `FileInfo` if we want to be able to reapply optimizations.
We can obsolete the other read method in that case to remain backwards compatible.

Project2015To2017/Reading/ProjectReader.cs
```diff
@@ -1,28 +1,48 @@
...
-			return Read(filePath, new Progress<string>(_ => { }));
+			progress = progress ?? new Progress<string>(_ => { });
+			return Read(new FileInfo(filePath), progress);
 		}
 
-		public Project Read(string filePath, IProgress<string> progress)
+		/// <summary>
+		/// Process-lifetime-long cache of loaded projects
+		/// <remarks>
+		/// It was supposed to be of <see cref="System.Runtime.Caching.ObjectCache"/> type, but that can still remove items from cache under memory pressure,
```
hvanbakel: This is absolute nitpicking but a cache is not supposed to give guarantees 😄

andrew-boyarshin:
@hvanbakel I will correct things you've mentioned and force push PR edits. Well, I use Dopamine (regularly) & PascalABC.NET (a couple of times) for XAML tests.
I am thinking of splitting this PR into more PRs too. Will see if I can do that simple enough.

Project2015To2017/Reading/ProjectReader.cs
andrew-boyarshin: Yeah, but that would incur reloading all project metadata that I would like to avoid. Seems like that is not as important as it did to me back then.

Project2015To2017/Reading/ProjectReader.cs
hvanbakel: I wasn't sure on the why of having the static field. I would keep it an instance still that you can initialize with either a file or a project definition. You could then by using obsolete attributes keep the existing contract so we're backwards compatible.

Or maybe I missed the intent of your change completely :smile:

andrew-boyarshin: Ok, you mean removing Instance field and adding constructors to initialize some `FileInfo Project` field? But then there is no point in caching, unless cache is made static. Is that ok?

hvanbakel:
I would expect that to be static, yes. That's around for the lifetime of
the application.

andrew-boyarshin:
@hvanbakel done. I also had to add .NET Standard 2.0 target for _System.Runtime.Caching_ NuGet package. On older targets (.NET Standard 1.3) old Dictionary solution is used. Only one is compiled at a time.

hvanbakel:
Merged as is

PR #142, opened 2018-08-01
CPS project reading support

andrew-boyarshin:
* Basic CPS reading support (VS2017 projects)
* Add `-z` command line switch (`--analyze`, allows CPS reading only for now)
* Minor fixes throughout the code

Most of the changes are related to XmlNamespace becoming Project instance-specific property.


Project2015To2017/ProjectConverter.cs
```diff
@@ -125,7 +133,7 @@ IProgress<string> progress
 			progress.Report("Solution parsing started.");
 			var solution = SolutionReader.Instance.Read(target, progress);
 
-			if (solution?.ProjectPaths == null)
+			if (solution.ProjectPaths == null)
```
andrew-boyarshin: It really can't be null (instance always created).

Project2015To2017/ProjectConverter.cs
```diff
@@ -15,6 +15,14 @@ public class ProjectConverter
 	{
 		private static IReadOnlyList<ITransformation> TransformationsToApply(ConversionOptions conversionOptions)
 		{
+			if (conversionOptions.AnalyzeOnly)
+			{
+				return new ITransformation[]
+				{
+					new FileTransformation(),
```
andrew-boyarshin: That's horrible, I believe we must rewrite `IncludeItems` mechanics completely, but for now that's a necessity.

Project2015To2017/Reading/AssemblyInfoReader.cs
```diff
@@ -29,6 +29,16 @@ public AssemblyAttributes Read(Project project, IProgress<string> progress)
 												}
 											)
 										   .Where(IsAssemblyInfoFile)
+										   .Where(x =>
+												{
+													if (x.Exists)
+													{
+														return true;
+													}
+													progress.Report($@"AssemblyInfo file '{x.FullName}' not found");
```
andrew-boyarshin: Prevents crash on [PascalABC.NET](https://github.com/pascalabcnet/pascalabcnet) subproject (missing AssemblyInfo file).

Project2015To2017/Reading/ProjectReader.cs
```diff
@@ -25,7 +25,8 @@ public ProjectReader(FileInfo projectFile, IProgress<string> progress = null)
 
 		public ProjectReader(string projectFilePath, IProgress<string> progress = null)
 		{
-			ProjectPath = new FileInfo(projectFilePath) ?? throw new ArgumentNullException(nameof(projectFilePath));
+			projectFilePath = projectFilePath ?? throw new ArgumentNullException(nameof(projectFilePath));
```
andrew-boyarshin: Shame on me. Previous PR had no null-check in fact.

Project2015To2017Tests/ProjectWriterTest.cs
```diff
@@ -333,7 +333,7 @@ void Deletion(FileSystemInfo info)
 
 				if (!File.Exists(file))
 				{
-					File.Create(file);
+					File.WriteAllBytes(file, new byte[0] {});
```
andrew-boyarshin: File.Create returns a Stream that is never flushed or closed... Very bug-prone.

Project2015To2017.Console/Options.cs
```diff
@@ -24,12 +24,16 @@ public class Options
 		[Option('o', "output-path", Default = false, HelpText = "Will not create a subfolder with the target framework in the output path")]
 		public bool NoTargetFrameworkToOutputPath { get; set; } = false;
 
+		[Option('z', "analyze-only", Default = false, HelpText = "Do only parsing and analysis steps, can be used even on already converted projects")]
+		public bool AnalyzeOnly { get; set; } = false;
```
hvanbakel: What's the difference between AnalyzerOnly and DryRun? They seem to be the same?

Project2015To2017.Console/Options.cs
```diff
@@ -24,12 +24,16 @@ public class Options
...
+
 		public ConversionOptions ConversionOptions
 			=> new ConversionOptions
 			{
 				KeepAssemblyInfo = AssemblyInfo,
 				TargetFrameworks = TargetFrameworks?.ToList(),
 				AppendTargetFrameworkToOutputPath = !NoTargetFrameworkToOutputPath,
+				AnalyzeOnly = AnalyzeOnly,
```
hvanbakel: This has nothing to do with this PR but shouldn't this assign DryRun too?

Project2015To2017.Console/Program.cs
```diff
@@ -51,6 +53,9 @@ private static void ConvertProject(Options options)
 				writer.Write(project, doBackup, progress);
 			}
 
+			System.Console.Out.Flush();
+
+			conversionOptions.AnalyzeOnly = true;
```
hvanbakel: I don't get this one, why do we need to set that here?

Project2015To2017.Console/Options.cs
andrew-boyarshin: DryRun only implies no writing will apply. AnalyzeOnly will also allow CPS project reading and analysis.
Analysis will also be ran as part of the conversion process, both before and after conversion (for future reference).

Project2015To2017/ProjectConverter.cs
hvanbakel: Yes, if only we had nullable reference types...

Project2015To2017.Console/Options.cs
andrew-boyarshin: I don't actually know, but in Program only original Options fields are used so no problem for now. I agree this is quite strange.

Project2015To2017.Console/Program.cs
andrew-boyarshin: Oh, that was part of next PR (analyzer). After writing we do full _read_ _&_ _analysis_ procedure again to account for changed locations of XML elements.

Project2015To2017Tests/UnsupportedProjectTypesTest.cs
```diff
@@ -4,6 +4,8 @@
 using System.Xml.Linq;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Project2015To2017;
+using Project2015To2017.Definition;
+using static Project2015To2017.Definition.Project;
```
hvanbakel: this line seems duplicate?

hvanbakel:
Looks good in general, I think we need to think about what a dry run and analyze only are and how they're different thought. It feels like they're the same thing.

Project2015To2017.Console/Options.cs
hvanbakel: doesn't that mean that we can collapse them into the same toggle? It doesn't make sense for dry run to just work for old projects?

Project2015To2017Tests/UnsupportedProjectTypesTest.cs
andrew-boyarshin: Actually my VS is down ( pending repairs :) ), so this PR was coupled in Sublime Text and `dotnet build`. But actually I don't know if `using static` implies reference of outer namespace.

andrew-boyarshin:
@hvanbakel I see your point. Maybe make `-z` imply `-n`? Or can adding some `-m` (modern) switch to allow CPS reading simplify matters?

hvanbakel:
I would say that if we want to support converted files too, we can just make that the default behavior, it will only work better. Then we don't need a switch for modern and can just have the analyze feature be toggled by the dry run flag?

Project2015To2017.Console/Options.cs
andrew-boyarshin: @hvanbakel uhm, suppose we might have post-conversion transformations. Like migration to .NET Core 3 (whatever) for users with CPS projects. That would complicate matters since we will want to have dry run for that scenario.

andrew-boyarshin:
@hvanbakel but writing support is not actually ready... unless we check for projectDefinition.IsModernProject and abort writing. Yeah, that would work. And in my implementation analysis is always done (with or without actual conversion) so the only actual behavior change is support for CPS reading. And that would be covered by making it default ON. And it will allow using dry-run for any kind of project. And that would allow us to drop this switch. Great idea. Will do right now unless there are any objections?

hvanbakel:
No objections, I think that makes it less confusing as to what the 2 options mean. Anything that we can do that is not changing actual files is fine :). Dry run just means don't alter anything which should be equivalent to your analyze option.

Project2015To2017.Console/Options.cs
andrew-boyarshin: That's no problem actually since there is no DryRun in `ConversionOptions`. :)

Project2015To2017Tests/UnsupportedProjectTypesTest.cs
andrew-boyarshin: No, that line is necessary for compilation (just checked it).

andrew-boyarshin:
@hvanbakel done.

Expected future PRs:
1. Analysis (base classes, ConsoleReporter, W001, W010, W020)
2. Diagnostic batch 1
3. Diagnostic batch 2
4. Indentation fixes & code cleanup
5. HTML reporter (for analysis, Razor templating)

hvanbakel:
And merged 👍 I did some basic cleanup for non used namespaces and removing those static usings.

Issue #141, opened 2018-07-27
Project analysis


andrew-boyarshin:
It seems we can implement some kind of analysis based on the information we gather. I've already implemented this part and I want more ideas for custom diagnostics.

* W001: Incompatible/Unsupported project type (Xamarin.*, WinForms)
* W002: Project file mentioned in solution is missing
* W010: $(Configurations)/$(Platforms) does not contain all configurations/platforms mentioned in project file
* W011: Conditional expression contains unsupported kind of subexpression (`and`, `not`)
* W012: File X that was not included before will be included after conversion thanks to wildcards
* W020: Wrong use of Microsoft.CSharp (usage on < net40, non-NuGet > net45)
* W021: Use System.* NuGet package
* W030: Consider migrating `pdbonly` & `full` DebugTypes to modern `portable` one.
* W031: You have MSBuild SDK version specified in your project file. A recommended approach would be using `global.json` for centralized version management.
* W032: You have old language version specified. Consider upgrading it to latest.
* W033: PCL profiles are obsolete. Consider migrating to .NET Standard.
* W034: Reference aliases are a feature of low support. Consider dropping their usage.

hvanbakel:
so for W012 I actually recieved a comment that that should not happen since it alters the state of the project while we could just exclude it. I think I would rather have a clean tree though (so keep the current behavior).

I think we should be warning for moving over to package reference and therefor losing the min max constraints on dependencies that you might have specified in a nuspec file. So far I haven't found a way to make VS set those (so I want my package to be compatible with [2.1, 3) so I'm not compatible with the next major.

andrew-boyarshin:
Diagnostics left to implement after #146:

[ ] W012
[ ] W013 (suggested by @hvanbakel)

andrew-boyarshin:
@hvanbakel I'm not sure about your PackageReference concerns. [This doc page](https://docs.microsoft.com/en-us/nuget/reference/package-versioning#examples) clearly shows it should work. Am I understanding your comment correctly?

hvanbakel:
@andrew-boyarshin it can support it. The fact that the references move from a nuspec to the project file means that every package you update means that you have to go back in and (manually) set the restrictions. Because this was a separate file your old restrictions weren't overwritten which it now does (as far as I know).

hvanbakel:
For W012, this is an interesting one, should we actually not create a node to exclude the file to keep the files included before/after the same?

[...]

PR #143, opened 2018-08-03
Basic analysis

andrew-boyarshin:
The foundation for future diagnostics.

#141 for reference.

### DIFF
--- a/Project2015To2017.Console/Program.cs
+++ b/Project2015To2017.Console/Program.cs
@@ -41,11 +41,7 @@ namespace Project2015To2017.Console
 
 			System.Console.Out.Flush();
 
-			var analysisOptions = new AnalysisOptions
-			{
-				ConversionOptions = conversionOptions
-			};
-			var analyzer = new Analyzer(analysisOptions);
+			var analyzer = new Analyzer();
 			foreach (var project in convertedProjects)
 			{
 				analyzer.Analyze(project);
--- a/Project2015To2017.Console/Program.cs
+++ b/Project2015To2017.Console/Program.cs
@@ -1,10 +1,11 @@
+using CommandLine;
+using Project2015To2017.Analysis;
+using Project2015To2017.Definition;
+using Project2015To2017.Reading;
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
-using CommandLine;
-using Project2015To2017.Definition;
-using Project2015To2017.Reading;
 
 namespace Project2015To2017.Console
 {
@@ -40,6 +41,16 @@ namespace Project2015To2017.Console
 
 			System.Console.Out.Flush();
 
+			var analysisOptions = new AnalysisOptions
+			{
+				ConversionOptions = conversionOptions
+			};
+			var analyzer = new Analyzer(analysisOptions);
+			foreach (var project in convertedProjects)
+			{
+				analyzer.Analyze(project);
+			}
+
 			if (options.DryRun)
 			{
 				return;
@@ -56,6 +67,7 @@ namespace Project2015To2017.Console
 					{
 						progressImpl.Report($"Skipping CPS project '{project.FilePath.Name}'...");
 					}
+
 					continue;
 				}
 
@@ -63,8 +75,30 @@ namespace Project2015To2017.Console
 			}
 
 			System.Console.Out.Flush();
-			
+
+			if (progress is IProgress<string> progressInterface)
+			{
+				progressInterface.Report("### Performing 2nd pass to analyze converted projects...");
+			}
+
 			ProjectReader.PurgeCache();
+			convertedProjects.Clear();
+
+			foreach (var file in options.Files)
+			{
+				var projects = ProjectConverter
+					.Convert(file, conversionOptions, progress)
+					.Where(x => x != null)
+					.ToList();
+				convertedProjects.AddRange(projects);
+			}
+
+			System.Console.Out.Flush();
+
+			foreach (var project in convertedProjects)
+			{
+				analyzer.Analyze(project);
+			}
 		}
 	}
 }
--- a/Project2015To2017/Analysis/AnalysisOptions.cs
+++ b/Project2015To2017/Analysis/AnalysisOptions.cs
@@ -1,0 +1,28 @@
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.IO;
+
+namespace Project2015To2017.Analysis
+{
+	public class AnalysisOptions
+	{
+		/// <summary>
+		/// Option set used for conversion, if any
+		/// </summary>
+		public ConversionOptions ConversionOptions { get; set; }
+		/// <summary>
+		/// Basement of all relative roots in output
+		/// </summary>
+		public DirectoryInfo RootDirectory { get; set; }
+		/// <summary>
+		/// Including ID of diagnostics in this list will make analyzer skip their execution and therefore output
+		/// </summary>
+		public ImmutableHashSet<DiagnosticBase> Diagnostics { get; }
+
+		public AnalysisOptions(IEnumerable<DiagnosticBase> diagnostics = null)
+		{
+			Diagnostics = (diagnostics ?? DiagnosticSet.AllDefault).ToImmutableHashSet();
+		}
+	}
+}
--- a/Project2015To2017/Analysis/AnalysisOptions.cs
+++ b/Project2015To2017/Analysis/AnalysisOptions.cs
@@ -1,26 +1,16 @@
-using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
-using System.IO;
 
 namespace Project2015To2017.Analysis
 {
 	public class AnalysisOptions
 	{
 		/// <summary>
-		/// Option set used for conversion, if any
-		/// </summary>
-		public ConversionOptions ConversionOptions { get; set; }
-		/// <summary>
-		/// Basement of all relative roots in output
-		/// </summary>
-		public DirectoryInfo RootDirectory { get; set; }
-		/// <summary>
 		/// Including ID of diagnostics in this list will make analyzer skip their execution and therefore output
 		/// </summary>
-		public ImmutableHashSet<DiagnosticBase> Diagnostics { get; }
+		public ImmutableHashSet<IDiagnostic> Diagnostics { get; }
 
-		public AnalysisOptions(IEnumerable<DiagnosticBase> diagnostics = null)
+		public AnalysisOptions(IEnumerable<IDiagnostic> diagnostics = null)
 		{
 			Diagnostics = (diagnostics ?? DiagnosticSet.AllDefault).ToImmutableHashSet();
 		}
--- a/Project2015To2017/Analysis/Analyzer.cs
+++ b/Project2015To2017/Analysis/Analyzer.cs
@@ -1,0 +1,100 @@
+using Project2015To2017.Analysis.Diagnostics;
+using Project2015To2017.Definition;
+using Project2015To2017.Reading;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Project2015To2017.Analysis
+{
+	public class Analyzer
+	{
+		/// <summary>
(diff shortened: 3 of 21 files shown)


Project2015To2017.Console/Program.cs
```diff
@@ -40,6 +41,16 @@ private static void ConvertProject(Options options)
 
 			System.Console.Out.Flush();
 
+			var analyzer = new Analyzer();
```
hvanbakel: Did you not make the options a constructor parameter on purpose?

Project2015To2017/Analysis/Analyzer.cs
```diff
@@ -0,0 +1,93 @@
...
+						continue;
+					}
+				}
+
+				if (options.SkipDiagnostics.Contains(diagnostic.Id))
+				{
+					continue;
+				}
+
+				diagnostic.Project = project;
```
hvanbakel: assigning these here means we cannot run analysis in parallel which could be a nice optimization for big solutions. Can we make the analyze call a side effect free function?

Project2015To2017/Analysis/Diagnostics/W001IllegalProjectTypeDiagnostic.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+
+namespace Project2015To2017.Analysis.Diagnostics
+{
+	public sealed class W001IllegalProjectTypeDiagnostic : DiagnosticBase
+	{
+		/// <inheritdoc />
+		public W001IllegalProjectTypeDiagnostic(Project project = null, AnalysisOptions options = null) : base(1, project, options)
```
hvanbakel: We need to use a constant for these analyzer ids so we can have a descriptive ‘skip diagnostics’ array that has the actual types that we opt out of.

Project2015To2017/Analysis/ReporterBase.cs
```diff
@@ -0,0 +1,41 @@
+using System.IO;
+using Project2015To2017.Transforms;
+
+namespace Project2015To2017.Analysis
+{
+	public abstract class ReporterBase
```
hvanbakel: I would rather use an interface here, this class doesn’t really do anything. You can achieve the same thing with an interface which is more flexible

hvanbakel:
@andrew-boyarshin this is an awesome change set 👍  I really like where this is going. You’re pulling all of the analysis and switching out of the actual conversion code and moving it into these analyzers. Again, if you want me to apply some of these fixes as part of the merge, I can do that too. Let me know.

Project2015To2017.Console/Program.cs
andrew-boyarshin: No, actually it's not important. I can change that.

Project2015To2017/Analysis/AnalysisOptions.cs
```diff
@@ -0,0 +1,31 @@
...
+		/// </summary>
+		public DirectoryInfo RootDirectory { get; set; }
+		/// <summary>
+		/// User reporter to use for output
+		/// </summary>
+		public ReporterBase Reporter { get; set; }
+		/// <summary>
+		/// Including ID of diagnostics in this list will make analyzer skip their execution and therefore output
+		/// </summary>
+		public IReadOnlyList<uint> SkipDiagnostics { get; set; } = new uint[0];
```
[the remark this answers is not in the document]
andrew-boyarshin: I've already mentioned my intention above, but even if it's rejected, why `HashSet`?

Project2015To2017/Analysis/Analyzer.cs
```diff
@@ -0,0 +1,93 @@
...
+using Project2015To2017.Reading;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Project2015To2017.Analysis
+{
+	public class Analyzer
+	{
+		public readonly IList<DiagnosticBase> Diagnostics = new List<DiagnosticBase>
```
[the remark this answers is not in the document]
hvanbakel: Exactly, move it into options with a default implementation.

Project2015To2017/Analysis/AnalysisOptions.cs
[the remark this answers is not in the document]
hvanbakel: Because that guarantees no dupes

Project2015To2017/Analysis/AnalysisOptions.cs
```diff
@@ -0,0 +1,31 @@
...
+		/// </summary>
+		public ConversionOptions ConversionOptions { get; set; }
+		/// <summary>
+		/// Basement of all relative roots in output
+		/// </summary>
+		public DirectoryInfo RootDirectory { get; set; }
+		/// <summary>
+		/// User reporter to use for output
+		/// </summary>
+		public ReporterBase Reporter { get; set; }
```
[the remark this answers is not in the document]
andrew-boyarshin: Reporting is how the analysis should do. But I see your point. I'll think about moving the property to Analyzer instead.

Project2015To2017/Analysis/Analyzer.cs
andrew-boyarshin: Fair point. But then it seems too many things might have to be passed to AnalyzeImpl as arguments. Project, Analyzer/Reporter. Well, if the number of these is not expected to grow — why not.

hvanbakel: I think it has to be
Project (what is analyzed)
Options (how to analyze)
Reporter (how to report)

That seems like a decent understandable set or parameters to me

Project2015To2017/Analysis/ReporterBase.cs
andrew-boyarshin: C# doesn't have default methods in interface, does it? And I'm not sure about sth being allowed to be a Reporter and something else at the same time.

hvanbakel: I think inheritance in implementation code (not in objects) is used to easily over just plain interfaces. Base classes have a tendency to become big. 

I’m not that eager on this change though. They are both valid solutions to the same problem

Project2015To2017/Analysis/AnalysisOptions.cs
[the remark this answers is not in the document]
andrew-boyarshin: Didn't think of the possibility. Is it just to reduce possible `Count`? HashSet is much heavier than List. We could just take only Distinct elements at the beginning of the analysis. And maybe sort them to add BinarySearch possibility.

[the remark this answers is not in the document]
hvanbakel: The list is probably not going to be big anyway so it’s probably not that important.

Project2015To2017/Analysis/Analyzer.cs
andrew-boyarshin: And now that I think of that, it would force the diagnostic implementers to pass Reporter to DiagnosticBase.Report. What about keeping Reporter and Options properties since they are common and immutable during the whole analysis time thus they won't affect parallelization?

hvanbakel: I don’t think the diagnostic should do the reporting, that’s the job of the analyzer, the diagnostic should just generate a diagnostic and return that to the analyzer.

Project2015To2017/Analysis/ReporterBase.cs
hvanbakel: If you keep this class, at least make it implement an interface so you can use that in other classes as to not tie it to a concrete type. In order to implement the behavior it doesn’t have to derive from `ReporterBase`.

Project2015To2017/Analysis/Analyzer.cs
andrew-boyarshin: I am familiar on the level of user. :) I haven't experienced the need to create my own diagnostic sets yet.
I believe the diagnostic in the end does as little job as possible. It really just finds violations and calls `Report`. Then the DiagnosticBase actually transforms the call into sth consumable by ConsoleReporter/UnitTestReporter/RazorReporter, the diagnostic impl doesn't have to know how it does that. But the DiagnosticBase doesn't do too much work either. It is still very generic. It calls to analysis property — Reporter that does all heavy lifting.
Analyzer is a pass through indeed. I intended it to be that way so that it only prepares everything and then bluntly calls each diagnostic for all projects. Nothing more, nothing less.
I believe we disagree on the responsibility to report, whether it belongs to Diagnostic or Analyzer. But I still don't understand how transferring the responsibility to Analyzer will simplify matters (from my point of view that would do otherwise).

hvanbakel: Well but if the reporter moves out of AnalyzerOptions, you’ll have to assign that as well. Maybe I can create a branch based on this one and make a proposal.

hvanbakel: I’ll see if I can create something tomorrow.

andrew-boyarshin: I will certainly take all comments into account and force push changes to this branch. Then you can create the proposal (in any repo, I added you as collaborator to my fork if you wish not to clutter this public one) and we will see if it would seem better for all sides of implementers.

Project2015To2017/Analysis/AnalysisOptions.cs
```diff
@@ -0,0 +1,28 @@
...
+		/// </summary>
+		public ConversionOptions ConversionOptions { get; set; }
+		/// <summary>
+		/// Basement of all relative roots in output
+		/// </summary>
+		public DirectoryInfo RootDirectory { get; set; }
+		/// <summary>
+		/// Including ID of diagnostics in this list will make analyzer skip their execution and therefore output
+		/// </summary>
+		public ImmutableHashSet<DiagnosticBase> Diagnostics { get; }
```
andrew-boyarshin: My mistake. Intended to be `ImmutableList` (no need for `ImmutableHashSet`, just sort and distinct in constructor). Next PR will fix that.

Project2015To2017/Analysis/Analyzer.cs
```diff
@@ -0,0 +1,100 @@
...
+		public void Analyze(Project project)
+		{
+			if (project == null)
+			{
+				throw new ArgumentNullException(nameof(project));
+			}
+
+			if (Options.RootDirectory == null)
+			{
+				Options.RootDirectory = project.Solution?.FilePath.Directory ?? project.FilePath.Directory;
```
andrew-boyarshin: Unfortunately, this is the mutation we have to agree with (no way to know the best RootDirectory in the current architecture before the analysis).

Project2015To2017/Analysis/DiagnosticBase.cs
```diff
@@ -0,0 +1,101 @@
...
+			{
+				sourceLine = (uint) elementOnLine.LineNumber;
+			}
+
+			Reporter.Report(DiagnosticCode, message, source, sourceLine);
+		}
+
+		public string DiagnosticCode => $"W{Id.ToString().PadLeft(3, '0')}";
+
+		public override int GetHashCode()
```
andrew-boyarshin: This and below are to significantly simplify matters for `HashSet`.

Project2015To2017/Analysis/DiagnosticSet.cs
```diff
@@ -0,0 +1,22 @@
+using System.Collections.Generic;
+using Project2015To2017.Analysis.Diagnostics;
+
+namespace Project2015To2017.Analysis
+{
+	public class DiagnosticSet : HashSet<DiagnosticBase>
```
andrew-boyarshin: Mutable until assignment to AnalysisOptions. This way the calling side can pick and remove any diagnostic before applying.

andrew-boyarshin:
I will be unavailable till ~11 am UTC, August 5th. I will be glad to see your proposal and/or comments on changes in this PR after that.

andrew-boyarshin:
Woah, didn't expect this merged... Wanted to push one more commit. Ok, next PR it is.

hvanbakel:
Sorry about that, I figured since you pulled it into the analysis branch that you were done

PR #150, opened 2018-08-13
Simplification refactorings + primitive property deduplication

### LINKED ISSUE #116: Removal of default assembly references should be moved from writer to a transform
This would then prevent an empty ItemGroup being written out when only default references were included

andrew-boyarshin:
* Move property simplification from ProjectPropertiesReader to transformation
* Create PropertyDeduplicationTransformation
* Move code from ProjectWriter to DefaultAssemblyReferenceRemovalTransformation (closes #116)
* Fix for CPS projects `$(Configurations)` and `$(Platforms)` reading
* Fix and move around tests to match changes
* Move Pre and Post transformations to ConversionOptions
* Simplify ProjectConverter methods


Project2015To2017Tests/XamlTransformationTest.cs
```diff
@@ -90,23 +90,6 @@ public async Task TransformsPresentationPages()
 
 			Assert.AreEqual(3, project.AdditionalPropertyGroups.Count);
 
-			Assert.IsNotNull(project.AdditionalPropertyGroups[0].Attribute("Condition"));
```
andrew-boyarshin: Does not relate to testing XAML transformation, moved to a new test for PropertySimplificationTransformation.

Project2015To2017Tests/ProjectWriterTest.cs
```diff
@@ -365,28 +366,6 @@ public void DeletedNonEmptyFolderIsNotProcessed()
 			CollectionAssert.AreEqual(new FileSystemInfo[0], actualDeletedFiles);
 		}
 
-		[TestMethod]
```
andrew-boyarshin: No more `ProjectWriter` responsibility. Moved to respective new test file.

Project2015To2017Tests/ProjectPropertiesReadTest.cs
```diff
@@ -642,142 +552,16 @@ public async Task ReadsUnknownConfigurations()
 			Assert.IsNull(project.AdditionalPropertyGroups[3].Attribute("Condition"));
 
 			var childrenDebug = project.AdditionalPropertyGroups[0].Elements().ToImmutableArray();
-			Assert.AreEqual(0, childrenDebug.Length);
+			Assert.AreEqual(2, childrenDebug.Length);
```
andrew-boyarshin: Project simplification is executed no more, test is duplicated in `PropertySimplificationTransformation` to check simplification.

Project2015To2017Tests/ProjectWriterTest.cs
```diff
@@ -142,11 +143,11 @@ public void SkipDelaySignNull()
 			var writer = new ProjectWriter();
 			var xmlNode = writer.CreateXml(new Project
 			{
-				DelaySign = null,
+				AdditionalPropertyGroups = new[] {new XElement("PropertyGroup")},
 				FilePath = new FileInfo("test.cs")
 			});
 
-			var delaySign = xmlNode.Element("PropertyGroup").Element("DelaySign");
+			var delaySign = xmlNode.Elements("PropertyGroup").Skip(1).First().Element("DelaySign");
```
andrew-boyarshin: The first `PropertyGroup` is the one created by `ProjectWriter`. `DelaySign` is from `AdditionalPropertyGroups`, so we need to skip the first node.

Project2015To2017/ProjectConverter.cs
```diff
@@ -11,9 +11,10 @@
 
 namespace Project2015To2017
 {
-	public class ProjectConverter
+	public static class ProjectConverter
```
hvanbakel: This is a breaking change, so I'm going to skip this one in the merge

Project2015To2017/ProjectConverter.cs
```diff
@@ -39,36 +43,14 @@ private static IReadOnlyList<ITransformation> TransformationsToApply(ConversionO
...
-		public static IEnumerable<Project> Convert(
-			string target,
-			IProgress<string> progress)
+		public static IEnumerable<Project> Convert(string target, IProgress<string> progress)
 		{
-			return Convert(target, new List<ITransformation>(), new List<ITransformation>(), progress);
+			return Convert(target, new ConversionOptions(), progress);
 		}
 
 		public static IEnumerable<Project> Convert(
```
hvanbakel: I'm going to make this backwards compatible decorating some Obsolete in there

hvanbakel:
I think this is a good change but there is no way to now make this break... @mungojam can you have a look as well, you might be one of the (few) people to be using the nuget package? I need to do a pass on the public API, if we're going to make a 2.0 I want to use it as a cleanup.

Project2015To2017/ProjectConverter.cs
andrew-boyarshin: Is it? I mean, do we care about binary compatibility or usage compatibility? The use is the same, no changes needed.

hvanbakel: You're right, it doesn't matter in this case.

Issue #154, opened 2018-08-16
2.0 issue rampdown roadmap


andrew-boyarshin:
* [x] #150 Simplification refactoring
* [x] #150 Property deduplication
* [x] Caching refactoring
* [x] #153 `IncludeItems` refactoring
* [x] #152 `AdditionalPropertyGroups` refactoring (fixed in #157)
* [x] `PackageReferenceTransformation` -> `TestProjectPackageReferenceTransformation`
* [x] #156 Add a better logging solution
* [x] #157 `PropertyGroup`s refactoring
* [x] #160 `PropertyGroups` refactoring (yeah, the 3rd one)
* [x] Verify public API surface class visibility and `sealed` possibility.
* [ ] Rewrite solution file parsing (to fix #161)
* [ ] Publish Release Candidate

hvanbakel:
I don't understand what you mean by the `PackageReferenceTransformation` item?

hvanbakel:
So I would add #156 as `add a better logging solution` to this list

andrew-boyarshin:
@hvanbakel — regarding PackageReferenceTransformation.
The only thing the transformation does is adding MSTest NuGet `PackageReference`s. I suggest renaming the transformation class to indicate that, instead of much more generic name. If additional PackageReference transformations would be required in the future, they l'd better be added as separate transformation.

hvanbakel:
I'll fix that now, and I'll put in a tentative interface for the logging which we can then implement internally.

andrew-boyarshin:
Test file is not renamed. :)

hvanbakel:
I'll do that too, and add a NetFX shim for
https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.ilogger?view=aspnetcore-2.1

andrew-boyarshin:
@hvanbakel there's LibLog library too, a different kind of abstraction, but I agree that we can get away with using Microsoft ILogger. We can change Console app later to use Serilog as ILogger backend if we feel like doing it.

hvanbakel:
Exactly, ILogger is a very nice abstraction in my opinion and it's the default abstraction so plenty of implementers out of the box. I'm working on putting that in right now and doing some other refactorings which should put is really close to a 2.0.

hvanbakel:
#156 should now also be addressed, I had to clone the console logger because there's no way to get around the newline for event ids... https://github.com/aspnet/Logging/issues/491

[...]

PR #157, opened 2018-08-20
Refactoring of PropertyGroups and bugfixes

andrew-boyarshin:
* Add `PrimaryPropertyGroup` property to `Project` model (stores the only unconditional `PropertyGroup`)
* Add `PrimaryUnconditionalPropertyTransformation` that takes care of the properties in unconditional `PropertyGroup`.
* Move primary property group logic to the newly created transformation
* Try harder at preserving original XML structure
* Make `ProjectPropertiesReader` non-static (like other readers), initialized with logger
* Remove unneeded arguments (like `XElement` when `Definition.Project` is already a parameter)
* Improve `FileTransformation` to be more correct
* Update `SdkExtrasVersion` to `MSBuild.Sdk.Extras/1.6.47`
* Reduce code duplication by moving helpers to Helpers (a surprise for sure, but a welcome one)
* Refactor `PropertySimplificationTransformation` to simplify the code and take other changes into account
* Allow CPS project writing
* Adapt tests to the changes
* Fix `FileTransformationTest` file indentation
* Remove the necessity of `FileTransformation` for CPS projects

Progress for #154.

### DIFF
--- a/Project2015To2017/Reading/ProjectPropertiesReader.cs
+++ b/Project2015To2017/Reading/ProjectPropertiesReader.cs
@@ -1,57 +1,67 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Text.RegularExpressions;
 using System.Xml.Linq;
+using Microsoft.Extensions.Logging;
 using Project2015To2017.Definition;
+using Project2015To2017.Transforms;
 
 namespace Project2015To2017.Reading
 {
-	public static class ProjectPropertiesReader
+	public class ProjectPropertiesReader
 	{
-		public static void PopulateProperties(Project project, XDocument projectXml)
+		private readonly ILogger _logger;
+
+		private static readonly string[] RemoveMSBuildImports =
 		{
-			var propertyGroups = projectXml.Element(project.XmlNamespace + "Project")
-				.Elements(project.XmlNamespace + "PropertyGroup")
-				.ToArray();
+			@"$(MSBuildToolsPath)\Microsoft.CSharp.targets",
+			@"$(MSBuildBinPath)\Microsoft.CSharp.targets",
+			@"$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props",
+		};
 
-			var unconditionalPropertyGroups = propertyGroups.Where(x => x.Attribute("Condition") == null).ToArray();
-			if (unconditionalPropertyGroups.Length == 0)
-			{
-				throw new NotSupportedException(
-					"No unconditional property group found. Cannot determine important properties like target framework and others.");
-			}
+		public ProjectPropertiesReader(ILogger logger)
+		{
+			_logger = logger ?? NoopLogger.Instance;
+		}
 
-			project.RootNamespace = unconditionalPropertyGroups.Elements(project.XmlNamespace + "RootNamespace")
-				.FirstOrDefault()?.Value;
-			project.AssemblyName = unconditionalPropertyGroups.Elements(project.XmlNamespace + "AssemblyName")
+		public void PopulateProperties(Project project)
+		{
+			ReadPropertyGroups(project);
+
+			project.RootNamespace = project.PrimaryPropertyGroup
+				.Elements(project.XmlNamespace + "RootNamespace")
 				.FirstOrDefault()
 				?.Value;
 
-			var targetFrameworkVersion = unconditionalPropertyGroups
+			project.AssemblyName = project.PrimaryPropertyGroup
+				.Elements(project.XmlNamespace + "AssemblyName")
+				.FirstOrDefault()
+				?.Value;
+
+			var targetFrameworkVersion = project.PrimaryPropertyGroup
 				.Elements(project.XmlNamespace + "TargetFrameworkVersion")
 				.FirstOrDefault();
 			if (targetFrameworkVersion?.Value != null)
 			{
 				project.TargetFrameworks.Add(ToTargetFramework(targetFrameworkVersion.Value));
-				targetFrameworkVersion.Remove();
 			}
 			else
 			{
-				var targetFrameworks = unconditionalPropertyGroups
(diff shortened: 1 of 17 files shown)


Project2015To2017/Reading/ProjectPropertiesReader.cs
```diff
@@ -1,180 +1,158 @@
...
-					platformSet.Add("AnyCPU");
-					needPlatforms = false;
+					platformSet.Add(platform);
 				}
 			}
-
-			if (needConfigurations || needPlatforms)
+			else
 			{
-				foreach (var x in projectXml.Descendants())
```
andrew-boyarshin: I decided that there is no need to traverse the XML tree. Neither did it match MSBuild behavior.

andrew-boyarshin: We emit diagnostic warning in mismatch cases.

Project2015To2017/Reading/ProjectReader.cs
```diff
@@ -15,6 +15,7 @@ public sealed class ProjectReader
 		private readonly Caching.IProjectCache projectCache;
 		private readonly NuSpecReader nuspecReader;
 		private readonly AssemblyInfoReader assemblyInfoReader;
+		private readonly ProjectPropertiesReader _projectPropertiesReader;
```
andrew-boyarshin: I intend later to rename other private instance fields to start with `_`, to match common C# codestyle conventions.

Project2015To2017/Transforms/FileTransformation.cs
```diff
@@ -173,19 +172,61 @@ private static bool KeepFileInclusion(XElement x, Project project)
 				return false;
 			}
 
+			if (linkElement == null)
```
andrew-boyarshin: Moved `Include` -> `Update` transformation to be **after** check if file is in project folder. Link is merely IDE UX guidance, it is not related to actual inclusion. If file is not in project folder, `Update` is not what we want..

Project2015To2017Tests/FileTransformationTest.cs
```diff
@@ -11,85 +9,85 @@
...
-            Assert.AreEqual(12, includeItems.Count(x => x.Name.LocalName == "Reference"));
-            Assert.AreEqual(2, includeItems.Count(x => x.Name.LocalName == "ProjectReference"));
-            Assert.AreEqual(11, includeItems.Count(x => x.Name.LocalName == "Compile"));
-			Assert.AreEqual(6, includeItems.Count(x => x.Name.LocalName == "Compile" && x.Attribute("Update") != null));
-			Assert.AreEqual(3, includeItems.Count(x => x.Name.LocalName == "Compile" && x.Attribute("Include") != null));
+			Assert.AreEqual(12, includeItems.Count(x => x.Name.LocalName == "Reference"));
+			Assert.AreEqual(2, includeItems.Count(x => x.Name.LocalName == "ProjectReference"));
+			Assert.AreEqual(11, includeItems.Count(x => x.Name.LocalName == "Compile"));
+			Assert.AreEqual(5, includeItems.Count(x => x.Name.LocalName == "Compile" && x.Attribute("Update") != null));
+			Assert.AreEqual(4, includeItems.Count(x => x.Name.LocalName == "Compile" && x.Attribute("Include") != null));
```
andrew-boyarshin: Reflects the abovementioned change in `if` statement ordering.

Project2015To2017Tests/FileTransformationTest.cs
```diff
@@ -11,85 +9,85 @@
...
 			Assert.AreEqual(2, includeItems.Count(x => x.Name.LocalName == "Compile" && x.Attribute("Remove") != null));
-			Assert.AreEqual(2, includeItems.Count(x => x.Name.LocalName == "EmbeddedResource")); // #73 inlcude things that are not ending in .resx
-            Assert.AreEqual(0, includeItems.Count(x => x.Name.LocalName == "Content"));
-            Assert.AreEqual(1, includeItems.Count(x => x.Name.LocalName == "None"));
-            Assert.AreEqual(0, includeItems.Count(x => x.Name.LocalName == "Analyzer"));
+			Assert.AreEqual(3, includeItems.Count(x => x.Name.LocalName == "EmbeddedResource")); // #73 include things that are not ending in .resx
```
andrew-boyarshin: Reflects fix of missing EmbeddedResource.

Project2015To2017/Reading/ProjectPropertiesReader.cs
hvanbakel: Lot of red here, I like it 👍

Project2015To2017/Reading/ProjectReader.cs
hvanbakel: That depends on who you ask, I never prefix with `_`. I think we should add a .editorconfig instead for consistency. It doesn't matter what you do as long as it's consistent.

hvanbakel: I'll make it back consistent in the merge let's spin up an issue of adding an editorconfig and getting everything in line.

Project2015To2017/Transforms/PropertySimplificationTransformation.cs
```diff
@@ -14,189 +14,185 @@ public sealed class PropertySimplificationTransformation : ITransformation
 	{
 		public void Transform(Project definition, ILogger logger)
 		{
-			FilterUnneededProperties(definition, definition.AdditionalPropertyGroups);
-		}
+			if (definition.PrimaryPropertyGroup == null)
+			{
+				throw new InvalidOperationException("Unable to do property simplification pass due to lack of primary PropertyGroup.");
```
hvanbakel: Not sure if this should be an exception though. Just continue if not found?

andrew-boyarshin: It was an exception before. And the situation is pretty weird. Even too weird for my liking.

PR #158, opened 2018-08-21
Proposal for 2.0

### LINKED ISSUE #116: Removal of default assembly references should be moved from writer to a transform

### LINKED ISSUE #151: IncludeItems is most inflexible
When we read `IncludeItems`, it contains all `ItemGroup`s.
`FileTransformation` then bluntly takes all actual items and now `Compile`, `Resource` and so on are the actual contents of `IncludeItems`.
There is a vast number of issues arising.

1. Overall changing the object type stored in variable.
2. Dependency on `FileTransformation` even for CPS projects.
3. Complete lack of `Condition` support for `ItemGroup`s.

Suggested way of tackling the issue:
1. `IncludeItems` -> `ItemGroups`, always store `ItemGroup`s.
2. Provide extension method/computed property to iterate items.
3. Huge refactoring of `FileTransformation`.
4. Modify all other callsites to work properly with data.

### LINKED ISSUE #152: Redundant unconditional PropertyGroup
We always create "main" PropertyGroup (for basic things like `TargetFrameworks`) and "additional" one for everything else. We need to merge these two before 2.0 release (obviously will require public API changes).

I intend to fix this one after #151.

### LINKED ISSUE #154: 2.0 issue rampdown roadmap

### LINKED ISSUE #155: app.config stays in project file
Firstly, you are famous, having been featured on [@shanselman's blog](https://www.hanselman.com/blog/UpgradingAnExistingNETProjectFilesToTheLeanNewCSPROJFormatFromNETCore.aspx)!

Watching through his youtube video on that post, he hits something I have noticed before but not reported or fixed which is that app.config gets left with a redundant entry in the project file.

I don't know what the rule should be for other .config files, but we could at least remove app.config ones.

### LINKED ISSUE #156: Implement line coloring for console output
Not entirely sure how to do this yet, but running the tool on a large project means that you might have a lot of lines being written that we should be classifying (not everything is an error).

hvanbakel:
Fixes #156 
Fixes #155 
Fixes #154 
Fixes #152 
Fixes #151 
Fixes #150 
Fixes #116


andrew-boyarshin:
See #159.

andrew-boyarshin:
I believe a good idea would be creating RC after I fix #159. This way @mungojam and others willing to test 2.0 will have an opportunity to adapt the codebase and we will be able to fix possible arising bugs.

hvanbakel:
That makes total sense, good suggestion

mungojam:
Very happy to test this, but it won't be till next week. It'd be interesting to try it against the open source projects that hanselman used it against and see how much better it does now. Not at a computer to find them at the moment

andrew-boyarshin:
A good testing suite would be:
* [Dopamine](https://github.com/digimezzo/Dopamine)
* [PascalABC.NET](https://github.com/pascalabcnet/pascalabcnet) - has lots of issues, code is structured like a mess
* [ConfuserEx](https://github.com/yck1509/ConfuserEx)
* [Force.com](https://github.com/developerforce/Force.com-Toolkit-for-NET) - thanks for the tip, @mungojam

mungojam:
I can't say I'm a fan of the `this.` everywhere. It feels like I'm reading some old Java. Isn't one of the points of using `_` in the field name that it's clear that you are referring to a field rather than a variable or parameter?

PR #160, opened 2018-08-21
One more refactoring of properties

### LINKED ISSUE #159: Multiple unconditional PropertyGroups
I have been wondering if we might be breaking something by merging all unconditional properties into one primary PropertyGroup. MSBuild doesn't support dependencies on values within one PropertyGroup, so user must separate variable value definition from its usage in e.g. condition of another property. The whole concept is weird and I'm not sure what we are to do.

andrew-boyarshin:
* `AdditionalPropertyGroups` -> `PropertyGroups`
* Move a lot of property handling logic to extension methods in `ProjectExtensions`
* Preserve all `PropertyGroup`s as-is
* Preserve `OutputType` and `TargetFramework` element locations
* Move common extension methods to library-global `Extensions` class
* New advanced property replacement logic
* Adapt tests to the changes

Fixes #159.


Project2015To2017/Project2015To2017.csproj
```diff
@@ -15,6 +15,7 @@
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
     <Version>1.8.1</Version>
     <AssemblyVersion>1.0.0.0</AssemblyVersion>
+    <LangVersion>7.3</LangVersion>
```
andrew-boyarshin: C# 7.2 requirement for `ref struct`, that allows more CLR runtime optimizations and prevents storing the find results anywhere (since the results can change very quickly there is no point in storing the results except to introduce bugs).

Project2015To2017/Transforms/PropertySimplificationTransformation.cs
```diff
@@ -95,7 +88,7 @@ public void Transform(Project definition, ILogger logger)
 					case "SccAuxPath" when !hasParentCondition && emptyValue:
 					case "SccProvider" when !hasParentCondition && emptyValue:
 					// Project properties set to defaults (Microsoft.NET.Sdk)
-					case "OutputType" when ValidateDefaultValue("Library"):
+					case "OutputType" when ValidateDefaultValue("library"):
```
andrew-boyarshin: That was an actual bug that was discovered by unit-tests after refactoring.

Project2015To2017/Analysis/Diagnostics/W021SystemNuGetPackagesDiagnostic.cs
```diff
@@ -143,7 +143,7 @@ public W021SystemNuGetPackagesDiagnostic() : base(21)
 			"System.IO.IsolatedStorage",
 		};
 
-		public StringComparison Comparison = ExtensionMethods.BestAvailableStringIgnoreCaseComparison;
+		public readonly StringComparison Comparison = Extensions.BestAvailableStringIgnoreCaseComparison;
```
hvanbakel: Can you make this a property?

Project2015To2017/Analysis/ExtensionMethods.cs
```diff
@@ -1,82 +0,0 @@
-using System;
```
hvanbakel: I had named all of these `ExtensionMethods` for consistency on the API. Probably not that big of a deal though since it’s all open source.

Project2015To2017/Reading/ProjectPropertiesReader.cs
```diff
@@ -161,32 +140,25 @@ private static (List<string>, List<string>) ReadConfigurationPlatformVariants(Pr
...
-					"No unconditional property group found. Cannot determine important properties like target framework and others.");
+				var _ = project.PrimaryPropertyGroup();
 			}
-
-			project.AdditionalPropertyGroups = conditional;
-
-			project.PrimaryPropertyGroup = unconditional[0];
-
-			foreach (var child in unconditional.Skip(1).SelectMany(x => x.Elements()))
+			catch
```
hvanbakel: We should be able to do this without using try catch for normal flow, right?

Project2015To2017/Analysis/Diagnostics/W021SystemNuGetPackagesDiagnostic.cs
andrew-boyarshin: The whole point was to cache the result, I dunno why it is public though. Will make `private` in next PR, what do you think?

hvanbakel: It’s a field so I would never want that in the public API, if we can make it private that solves it.

Project2015To2017/Reading/ProjectPropertiesReader.cs
andrew-boyarshin: An exception is already thrown (`First` instead of `FirstOrDefault`), this is just to change the exception kind and message.

hvanbakel: Right, but in that case catch a more specific exception, these very generic catch statements might hide a lot more than you initially intended

Project2015To2017/Analysis/ExtensionMethods.cs
andrew-boyarshin: Now there are:
* `P15To17.Analysis.AnalysisExtensions`
* `P15To17.Definition.ProjectExtensions`
* `P15To17.Extensions`

Seems pretty consistent and describing the purpose.

hvanbakel: Yeah, maybe went a bit overboard there. The initial comment here was more of a reason why I did it rather than it needs to be changed.

Project2015To2017/Reading/ProjectPropertiesReader.cs
andrew-boyarshin: Am I to make changes in this PR or can we delay that to the next PR?

Project2015To2017/Analysis/ExtensionMethods.cs
andrew-boyarshin: Oh, I see.

Project2015To2017/Reading/ProjectPropertiesReader.cs
hvanbakel: Whatever is most convenient to you

andrew-boyarshin: Next PR it is (unable to work on that now).

hvanbakel:
No worries, I'll apply the fixes myself.

Issue #162, opened 2018-08-21
Global tool package and exe are named differently


mungojam:
One comment Scott Hanselman makes on [the blog](https://www.hanselman.com/blog/UpgradingAnExistingNETProjectFilesToTheLeanNewCSPROJFormatFromNETCore.aspx) and video is that he has to type a different command to run the global tool than what it is called. I haven't used many global tools to know how common that is but it would make sense to try and align them.

hvanbakel:
I heard that, thought about and I don’t think I want to be doing that. If I could go back and rename the entire repository and all the projects, then that might be worthwhile but I wanted something that wasn’t using any capitals and still relatively short. 

I don’t think it’s that much of a problem, yes you need to look at the page which explained how to install it, that’s a one time thing though.

mungojam:
Perhaps we could add a note or link to the text that appears on
https://www.nuget.org/packages/Project2015To2017.Cli

I don't think we should assume that people know to look at the github
readme and some companies annoyingly block github.com

hvanbakel:
We should absolutely be able to do that. Let’s reopen this one and make
that the ‘fix’.

mungojam:
A couple more observations, it actually tells you when installing what the command is:

> C:\Users\mark\source\repos\CsprojToVs2017>dotnet tool install --global Project2015To2017.Cli
**You can invoke the tool using the following command: csproj-to-2017**
Tool 'project2015to2017.cli' (version '1.8.1') was successfully installed.

I can see the difficulty you had as I don't think hyphens are allowed in the package name. I see some other global tools like NuKeeper use PascalCase for their main exe:

>C:\Users\mark\source\repos\CsprojToVs2017>dotnet tool install --global NuKeeper
You can invoke the tool using the following command: **NuKeeper**
Tool 'nukeeper' (version '0.6.3') was successfully installed.

so you could go for CsProjTo2017 for the package name and tool exe. But your call

hvanbakel:
Done, see here:
https://www.nuget.org/packages/Project2015To2017.Cli

hvanbakel:
I might look into aliases for the tool later but for now this works I guess.

andrew-boyarshin:
@hvanbakel I like your idea mentioned in #169. I'd like to elaborate on the way I see it would be even better.

```batch
> dotnet migrate-2017
(usage help)
> dotnet migrate-2017 evaluate
[...]

PR #169, opened 2018-08-22
Vbproj support

### LINKED ISSUE #74: Vbproj support?
Do you plan to support vb projects?
There should not be a lot of difference I guess..

hvanbakel:
Fixes #74.

I don't know if either of you guys has a bunch of vb or F# projects lying around. I don't which was why I never wanted to put it into master.

I guess we can commit it as 'experimental'...

We would probably want to create a separate tool, maybe as dotnet-migrate-2017 so you could do:
`dotnet migrate-2017` 
or something similar. That might also address #162

### DIFF
--- a/Project2015To2017/ProjectConverter.cs
+++ b/Project2015To2017/ProjectConverter.cs
@@ -91,14 +91,14 @@ namespace Project2015To2017
 				yield break;
 			}
 
+			var projectsProcessed = 0;
 			// Process all csprojs found in given directory
 			foreach (var mapping in ProjectFileMappings)
 			{
 				var projectFiles = Directory.EnumerateFiles(target, "*" + mapping.Key, SearchOption.AllDirectories).ToArray();
 				if (projectFiles.Length == 0)
 				{
-					this.logger.LogCritical("Please specify a project file.");
-					yield break;
+					continue;
 				}
 
 				if (projectFiles.Length > 1)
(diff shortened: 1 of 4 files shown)


hvanbakel:
disclaimer, this is the first time ever I did a rebase, so hopefully that was done correctly 😉

mungojam:
I've not used vb.net before either or much fsharp.

For rebasing like this I tend to just use VS and then you wouldn't change the commit message to mention rebasing. The idea of this type of rebasing is to make it look like history was linear, like you just started on VB projects support today, so first commit would be 'add vbproj support'

Project2015To2017/ProjectConverter.cs
```diff
@@ -52,25 +59,21 @@ public IEnumerable<Project> Convert(string target)
...
-						yield return this.ProcessFile(file, null);
-						break;
-
-					default:
+					foreach (var project in ConvertSolution(target))
+					{
+						yield return project;
+					}
+				}
+				else if (ProjectFileMappings.TryGetValue(extension, out var fileExtension))
```
andrew-boyarshin: `switch` (with `when`) would still seem nicer, but that's not a real issue to be fixed.

hvanbakel:
Thanks for the review, I need to make more use of the fancy new features
indeed.

I won't merge prior to 2.0 being out though.

Issue #165, opened 2018-08-21
Collection of issues from Force.com toolkit example


mungojam:
I've now run the conversion of the SalesForce example (from shanselman blog) using the [unconditional branch](https://github.com/andrew-boyarshin/CsprojToVs2017/tree/unconditional). There are a few things left in the project file that I thought I had sorted, but also some that might be easy to fix.

You can see all the converted project files in [my fork](https://github.com/mungojam/Force.com-Toolkit-for-NET/blob/converter-v2/src)

Just going to list these here to get opinions on whether they are fixable in the current work or should be issues for future work:

[ChatterToolkitForNet.csproj](https://github.com/mungojam/Force.com-Toolkit-for-NET/blob/converter-v2/src/ChatterToolkitForNET/ChatterToolkitForNET.csproj)

Could we remove this?:
```xml
    <MinimumVisualStudioVersion>11.0</MinimumVisualStudioVersion>
```

I thought we can remove project GUIDs?:
```xml
    <ProjectGuid>{011B1F31-45AA-4CD1-BB97-F4A508408365}</ProjectGuid>
```

We could provide an option to automatically convert to netstandard, or to remove portable project settings if the consumer chooses to upgrade to netstandard
```xml
    <TargetFrameworkProfile>Profile111</TargetFrameworkProfile>
    <ProjectTypeGuids>{786C830F-07A1-408B-BD7F-6EE04809D6DB};{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}</ProjectTypeGuids>
```

This feels like something auto-generated, so can we detect and remove it?
```xml
    <SolutionDir Condition="$(SolutionDir) == '' Or $(SolutionDir) == '*Undefined*'">..\</SolutionDir>
```

Pretty sure this is default nowadays
```xml
    <RestorePackages>true</RestorePackages>
```

Not sure on these, but I would have thought we can generally ditch them. Maybe we need a safe mode and an eager mode!
```xml
    <ProductVersion>12.0.0</ProductVersion>
    <SchemaVersion>2.0</SchemaVersion>
    <OutputPath>bin\$(Configuration)\</OutputPath>
```

These are now the ones that I thought I had set to be removed if they were the same because msbuild and VS only need one of them if they all match. I had worked out which one, but maybe hadn't put it into a test.

```xml
    <RootNamespace>Salesforce.Chatter</RootNamespace>
    <AssemblyName>Salesforce.Chatter</AssemblyName>
[...]

PR #170, opened 2018-08-22
Last bit of fixes before 2.0 release

### LINKED ISSUE #161: Error on macos: Project file not found at
It seems that the different slashes confuse it:

```
csproj-to-2017 ./cryptoportfolio.sln 
Solution parsing started.
Project found: cryptoportfolio\cryptoportfolio.csproj
Project file not found at: /Users/rob/Dropbox/src/cryptoportfolio/cryptoportfolio\cryptoportfolio.csproj

```
Project file is there, but it should use path /Users/rob/Dropbox/src/cryptoportfolio/cryptoportfolio/cryptoportfolio.csproj

In the SLN the proj is defined as 
`Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "cryptoportfolio", "cryptoportfolio\cryptoportfolio.csproj", "{3C579E5C-ABF9-4286-AAF3-8230FB37BEB7}"
`

andrew-boyarshin:
* API removal: Remove `RootNamespace` and `AssemblyName` properties from model (it was RC right? one tiny change)
* Extended `ProjectReference` model (and added filling new properties to both `SolutionReader` and `ProjectReader`)
* Use new APIs in diagnostics W001, W010, W034
* Fix AssemblyInfo detection when file is missing (e.g. generated during build from template)
* Extend `ConditionEvaluator` APIs to allow more efficient queries
* Recreate relative path fix set for Unix-like OS closely based on MSBuild sources (works only on non-Windows platforms)
* Rewrite `SolutionReader` to use MSBuild routine for parsing _Project_ definition lines
* Add `EmptyGroupRemoveTransformation` (obvious designation, fixes regression in one of my latest PRs)
* Greatly improve property simplification per issue #165
* Fix regression in one of my latest PRs: applying full set of transformations on CPS projects. Sigh.
* Adapt tests to changes

Closes #161.

### DIFF
--- a/Project2015To2017/ProjectConverter.cs
+++ b/Project2015To2017/ProjectConverter.cs
@@ -18,13 +18,24 @@ namespace Project2015To2017
 		private readonly ConversionOptions conversionOptions;
 		private readonly ProjectReader projectReader;
 
-		private static IReadOnlyCollection<ITransformation> TransformationsToApply(ConversionOptions conversionOptions)
+		private static IReadOnlyCollection<ITransformation>
+			TransformationsToApply(ConversionOptions conversionOptions, bool modernProject)
 		{
+			var targetFrameworkTransformation = new TargetFrameworkTransformation(
+				conversionOptions.TargetFrameworks,
+				conversionOptions.AppendTargetFrameworkToOutputPath);
+
+			if (modernProject)
+			{
+				return new ITransformation[]
+				{
+					targetFrameworkTransformation
+				};
+			}
+
 			return new ITransformation[]
 			{
-				new TargetFrameworkTransformation(
-					conversionOptions.TargetFrameworks,
-					conversionOptions.AppendTargetFrameworkToOutputPath),
+				targetFrameworkTransformation,
 				new PropertySimplificationTransformation(),
 				new PropertyDeduplicationTransformation(),
 				new TestProjectPackageReferenceTransformation(),
@@ -37,6 +48,7 @@ namespace Project2015To2017
 				new AssemblyAttributeTransformation(conversionOptions.KeepAssemblyInfo),
 				new XamlPagesTransformation(),
 				new PrimaryUnconditionalPropertyTransformation(),
+				new EmptyGroupRemoveTransformation(),
 			};
 		}
 
@@ -120,21 +132,20 @@ namespace Project2015To2017
 				yield break;
 			}
 
-			foreach (var projectPath in solution.ProjectPaths)
+			foreach (var projectReference in solution.ProjectPaths)
 			{
-				this.logger.LogInformation("Project found: " + projectPath.Include);
-				if (!projectPath.ProjectFile.Exists)
+				this.logger.LogInformation("Project found: " + projectReference.Include);
+				if (!projectReference.ProjectFile.Exists)
 				{
-					this.logger.LogError("Project file not found at: " + projectPath.ProjectFile.FullName);
+					this.logger.LogError("Project file not found at: " + projectReference.ProjectFile.FullName);
+					continue;
 				}
-				else
-				{
-					yield return this.ProcessFile(projectPath.ProjectFile, solution);
-				}
+
+				yield return this.ProcessFile(projectReference.ProjectFile, solution, projectReference);
 			}
 		}
 
-		private Project ProcessFile(FileInfo file, Solution solution)
+		private Project ProcessFile(FileInfo file, Solution solution, ProjectReference reference = null)
 		{
 			if (!Validate(file, this.logger))
 			{
@@ -147,6 +158,11 @@ namespace Project2015To2017
 				return null;
 			}
 
+			if (reference?.ProjectName != null)
+			{
+				project.ProjectName = reference.ProjectName;
+			}
+
 			project.Solution = solution;
 
 			foreach (var transform in this.conversionOptions.PreDefaultTransforms)
@@ -154,7 +170,7 @@ namespace Project2015To2017
 				transform.Transform(project, this.logger);
 			}
 
-			foreach (var transform in TransformationsToApply(this.conversionOptions))
+			foreach (var transform in TransformationsToApply(this.conversionOptions, project.IsModernProject))
 			{
 				transform.Transform(project, this.logger);
 			}
--- a/Project2015To2017/Reading/ProjectPropertiesReader.cs
+++ b/Project2015To2017/Reading/ProjectPropertiesReader.cs
@@ -28,9 +28,6 @@ namespace Project2015To2017.Reading
 		{
 			ReadPropertyGroups(project);
 
-			project.RootNamespace = project.Property("RootNamespace", tryConditional: true)?.Value;
-			project.AssemblyName = project.Property("AssemblyName", tryConditional: true)?.Value;
-
 			var targetFrameworkVersion = project.Property("TargetFrameworkVersion")?.Value;
 			if (targetFrameworkVersion != null)
 			{
@@ -145,7 +142,7 @@ namespace Project2015To2017.Reading
 				.Split(new[] {';'}, StringSplitOptions.RemoveEmptyEntries);
 		}
 
-		private static void ReadPropertyGroups(Project project)
+		internal static void ReadPropertyGroups(Project project)
 		{
 			project.PropertyGroups = project.ProjectDocument.Root
 				.Elements(project.XmlNamespace + "PropertyGroup")
--- a/Project2015To2017/Reading/ProjectReader.cs
+++ b/Project2015To2017/Reading/ProjectReader.cs
@@ -150,13 +150,21 @@ namespace Project2015To2017.Reading
 		{
 			foreach (var reference in projectDefinition.ProjectReferences)
 			{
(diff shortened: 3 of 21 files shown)


Project2015To2017/Reading/ProjectPropertiesReader.cs
```diff
@@ -145,7 +142,7 @@ string[] ParseFromProperty(string name) => project.Property(name)
 				.Split(new[] {';'}, StringSplitOptions.RemoveEmptyEntries);
 		}
 
-		private static void ReadPropertyGroups(Project project)
+		internal static void ReadPropertyGroups(Project project)
```
andrew-boyarshin: For accessibility in tests.

Project2015To2017/Reading/SolutionReader.cs
```diff
@@ -11,35 +11,74 @@ public sealed class SolutionReader
 	{
 		public static readonly SolutionReader Instance = new SolutionReader();
 
+		// An example of a project line looks like this:
+		//  Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "ClassLibrary1", "ClassLibrary1\ClassLibrary1.csproj", "{05A5AD00-71B5-4612-AF2F-9EA9121C4111}"
+		private static readonly Lazy<Regex> crackProjectLine = new Lazy<Regex>(
```
andrew-boyarshin: I still believe `Lazy` is a good idea. The overhead is actually small (I looked at `Lazy` sources), but improves cases when there is no need to **compile** heavy Regex, like when only projects are specified.

Project2015To2017/ProjectConverter.cs
```diff
@@ -18,13 +18,24 @@ public sealed class ProjectConverter
...
 		{
+			var targetFrameworkTransformation = new TargetFrameworkTransformation(
+				conversionOptions.TargetFrameworks,
+				conversionOptions.AppendTargetFrameworkToOutputPath);
+
+			if (modernProject)
+			{
+				return new ITransformation[]
+				{
+					targetFrameworkTransformation
```
mungojam: Could we keep some of the other transforms? They might apply if somebody had used an older version of the converter that didn't have all of the current simplifications. Not needed for this release though

Project2015To2017/Reading/SolutionReader.cs
mungojam: I think in those cases it won't ever initialise SolutionReader including the static fields, but could be wrong

Project2015To2017/Reading/SolutionReader.cs
```diff
@@ -11,35 +11,74 @@ public sealed class SolutionReader
...
+				+ "\"(?<RELATIVEPATH>.*)\""
+				+ "\\s*,\\s*" // Any amount of whitespace plus "," plus any amount of whitespace
+				+ "\"(?<PROJECTGUID>.*)\""
+				+ "$", // End-of-line
+				RegexOptions.Compiled
+			)
+		);
+
+		private const string vbProjectGuid = "{F184B08F-C81C-45F6-A57F-5ABD9991F28F}";
+		private const string csProjectGuid = "{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}";
```
mungojam: are you sure this covers all generations of Visual Studio and things like test projects? What about WPF? I had a feeling that there were more possible values it could take.

Project2015To2017/Transforms/EmptyGroupRemoveTransformation.cs
```diff
@@ -0,0 +1,30 @@
...
+		public void Transform(Project definition, ILogger logger)
+		{
+			definition.PropertyGroups = FilterNonEmpty(definition.PropertyGroups);
+			definition.ItemGroups = FilterNonEmpty(definition.ItemGroups).ToList();
+		}
+
+		private static IReadOnlyList<XElement> FilterNonEmpty(IEnumerable<XElement> groups)
+		{
+			var (keep, remove) = groups
+				.Split(x => x.HasElements
```
mungojam: Do you need to filter out comments here too? Here's an example from the current converter which produces [this csproj](https://github.com/mungojam/Force.com-Toolkit-for-NET/blob/with-pr-167/src/ForceToolkitForNET/ForceToolkitForNET.csproj) with:
```xml
  <ItemGroup>
    <!-- A reference to the entire .NET Framework is automatically included -->
  </ItemGroup>
```

Project2015To2017/Transforms/NuGetPackageTransformation.cs
```diff
@@ -76,10 +77,8 @@ private string PopulatePlaceHolder(string placeHolder, string nuSpecValue, strin
 			{
 				return assemblyAttributeValue ?? nuSpecValue;
 			}
-			else
-			{
-				return nuSpecValue;
-			}
+
+			return nuSpecValue;
```
mungojam: I guess we have different styles here. This is one resharper warning we disable at work because we prefer the clarity of if/else

mungojam:
The new solution parser is great, but it's getting confused by solution folders. I ran it against [this solution](https://github.com/mungojam/Force.com-Toolkit-for-NET/blob/with-stripped-projects/src/ForceToolkitForNET.sln) and it warned about the two solution folders:

> warn Unsupported project[Tests] type {2150E333-8FDC-42A3-9474-1A3956D46DE8}
warn Unsupported project[.nuget] type {2150E333-8FDC-42A3-9474-1A3956D46DE8}

Project2015To2017/Transforms/EmptyGroupRemoveTransformation.cs
mungojam: Actually, it's [not happening anymore](https://github.com/mungojam/Force.com-Toolkit-for-NET/blob/rc-fixes/src/ForceToolkitForNET/ForceToolkitForNET.csproj) so probably fine

Project2015To2017/Reading/ProjectReader.cs
```diff
@@ -150,13 +150,21 @@ private static void ProcessProjectReferences(Project projectDefinition)
...
+				var projectDirectory = projectDefinition.FilePath.DirectoryName;
+				var adjustedPath = Extensions.MaybeAdjustFilePath(path, projectDirectory);
+				reference.ProjectFile = new FileInfo(Path.Combine(projectDirectory, adjustedPath));
 			}
 		}
 
 		private FileInfo FindPackagesConfigFile(FileInfo projectFile)
 		{
-			var packagesConfig = new FileInfo(Path.Combine(projectFile.Directory.FullName, "packages.config"));
+			var packagesConfig = new FileInfo(Path.Combine(projectFile.DirectoryName, "packages.config"));
```
hvanbakel: This doesn't work if your directory is nested, right? Fullname took the entire path, this makes the assumption of being one folder deep?

Project2015To2017/Reading/SolutionReader.cs
andrew-boyarshin: I'm afraid we can't guarantee that. The first time `Instance` or any other field is accessed, all static fields are initialized. Imagine we add sometime later a new static field to SolutionReader and access that in ProjectConverter unconditionally, then regex will always be initialized and compiled. MSBuild folks didn't do it out of pure will to have _perfect_ code.

Project2015To2017/ProjectConverter.cs
andrew-boyarshin: That's a matter for different PR. We shouldn't change CPS project files at all when we read and write them, unless any additional transformations are passed in `ConversionOptions`.

Project2015To2017/Reading/SolutionReader.cs
andrew-boyarshin: These are not the ProjectTypeGuids from project file. These are specific solution Guids, that cover all managed languages supported by latest MSBuild but J# and C++/CLI. This was taken directly from MSBuild sources.

Project2015To2017/Reading/ProjectReader.cs
andrew-boyarshin: I'm afraid it's incorrect. [DirectoryName reference](https://docs.microsoft.com/en-us/dotnet/api/system.io.fileinfo.directoryname?redirectedfrom=MSDN&view=netframework-4.7.2#System_IO_FileInfo_DirectoryName).
`Directory` property internally calls `DirectoryName`, checks `string.IsNullOrEmpty`, and creates `DirectoryInfo` instance from that string.

andrew-boyarshin:
@hvanbakel please, reconsider your review decision taking my comments in consideration.

andrew-boyarshin:
@mungojam it doesn't get confused by folder, it just outputs warning and ignores this "folder" project. Maybe we indeed could ignore warning for the respective TypeGuid.

Project2015To2017/Reading/ProjectReader.cs
hvanbakel: Interesting, you’re right I was thinking of `DirectoryInfo.Name`

mungojam:
> Maybe we indeed could ignore warning for the respective TypeGuid

Yes please let's do that. Current behaviour is confusing

PR #174, opened 2018-08-22
Refactored ITransformation to not include logger for 2.0

hvanbakel:
Putting `ILogger` as a parameter on the `ITransformation` call never really made sense to me so now that's refactored. Since that's a change to the public API it would need to go in prior to 2.0 non RC.

### DIFF
--- a/Project2015To2017/ProjectConverter.cs
+++ b/Project2015To2017/ProjectConverter.cs
@@ -17,34 +17,14 @@ namespace Project2015To2017
 		private readonly ILogger logger;
 		private readonly ConversionOptions conversionOptions;
 		private readonly ProjectReader projectReader;
-
-		private static IReadOnlyCollection<ITransformation> TransformationsToApply(ConversionOptions conversionOptions)
-		{
-			return new ITransformation[]
-			{
-				new TargetFrameworkTransformation(
-					conversionOptions.TargetFrameworks,
-					conversionOptions.AppendTargetFrameworkToOutputPath),
-				new PropertySimplificationTransformation(),
-				new PropertyDeduplicationTransformation(),
-				new TestProjectPackageReferenceTransformation(),
-				new AssemblyReferenceTransformation(),
-				new RemovePackageAssemblyReferencesTransformation(),
-				new DefaultAssemblyReferenceRemovalTransformation(),
-				new RemovePackageImportsTransformation(),
-				new FileTransformation(),
-				new NugetPackageTransformation(),
-				new AssemblyAttributeTransformation(conversionOptions.KeepAssemblyInfo),
-				new XamlPagesTransformation(),
-				new PrimaryUnconditionalPropertyTransformation(),
-			};
-		}
+		private readonly IReadOnlyCollection<ITransformation> transformationsToApply;
 
 		public ProjectConverter(ILogger logger, ConversionOptions conversionOptions = null)
 		{
 			this.logger = logger;
 			this.conversionOptions = conversionOptions ?? new ConversionOptions();
 			this.projectReader = new ProjectReader(logger, this.conversionOptions);
+			this.transformationsToApply = TransformationsToApply(this.conversionOptions, this.logger);
 		}
 
 		public IEnumerable<Project> Convert(string target)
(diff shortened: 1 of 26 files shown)


Project2015To2017/ProjectConverter.cs
```diff
@@ -177,5 +157,27 @@ internal static bool Validate(FileInfo file, ILogger logger)
 			logger.LogError($"File {file.FullName} could not be found.");
 			return false;
 		}
+
+		private static IReadOnlyCollection<ITransformation> TransformationsToApply(ConversionOptions conversionOptions, ILogger logger)
+		{
+			return new ITransformation[]
```
andrew-boyarshin: Hmm. That would cause merge conflicts with #170.

hvanbakel: Yes there will be merge conflicts.

hvanbakel:
So the thing is that the logger aspect is no no longer a requirement of building a transformation. The constructor is not part of the interface so you can just use anything you want in there. Some use a logger some don’t. This also means that if you create a transformation that doesn’t need a logger than you can just instantiate that without one.

I would not introduce a base class for precisely that reason, not all transforms need a logger. It’s like with what is in #170 where you pass in IsModerProject only to 1 constructor. I don’t see how that’s an issue?

hvanbakel:
I’m referring to this:
https://github.com/hvanbakel/CsprojToVs2017/pull/170/files#diff-066798b475d8298c4f6db3657eaf4324
The FileTransformation uses some arbitrary constructor arguments, I don’t see how that’s different from passing in a logger.

Logger should be some ambient construct really but that’s gold plating it...

andrew-boyarshin:
Not that I like the `TargetFrameworkTransformation` approach, I was not the one who implemented it that way.
If only C# supported default interface method implementations — that would simplify matters. Or constructor signatures in interface (ILogger parameter could be optional).
Transformation is not obliged indeed to use logging abilities, but is it good from library perspective to not be aware of how precisely to initialize any transformation?

hvanbakel:
Ok but the `ITransformation` interface is the generalization over all transformations. It describes the contract you need to adhere to in order to write one that we can execute. Things like `AttachLogger` could be put in a different interface and then we could solve it that way but that's weird to me for something that is just a dependency for the type. Dependencies belong in the constructor and some types are dependent upon a logger and some are not, that's why they're there.

andrew-boyarshin:
Ok, a better way might indeed be additional `ILoggingTransformation` interface. This way it might make it more explicit, that this Transformation uses logging and should be initialized with it.

hvanbakel:
@andrew-boyarshin I would agree with you if I would want to specify different loggers in different cases. In this case though, this is initialized only once and all cases use the same one. Unless there's a reason why that might be different, I don't see an issue with using the constructor for this as it's a dependency of the type not some 'optional' thing that you can bolt on if you want.

andrew-boyarshin:
Well, I guess we'll have to agree to disagree on this point. I withdraw my objection.

Project2015To2017/Transforms/PropertySimplificationTransformation.cs
```diff
@@ -13,6 +13,12 @@ namespace Project2015To2017.Transforms
 {
 	public sealed class PropertySimplificationTransformation : ITransformation
 	{
+		private static readonly string[] IgnoreProjectNameValues =
```
andrew-boyarshin: Uhm... What is that?

hvanbakel:
On the upside we both agree that it doesn't belong on `ITransformation`.

Aside from the argument do you have any code examples where you use the above. This code style comes from the use of Dependency Injection where all the dependencies go in the constructor. It sounds to me like you have a different paradigm which is always good to look at and consider.

Project2015To2017/Transforms/PropertySimplificationTransformation.cs
hvanbakel: I'll have a look later, a merge issue maybe...

andrew-boyarshin: It is from #170, matching current code, but still present here and not causing merge conflicts. What?

andrew-boyarshin: @hvanbakel yeah, that looks like a merge issue. Once PR is cleaned up of this unneeded change — count my approval in effect.

andrew-boyarshin:
@hvanbakel honestly speaking, I'm not on a good terms with DI, in fact, I've never used one except to learn the technique. I like the idea of passing `ILogger` in general, yet I believe the consistency is vital too, so if similar classes implementing one interface take unstable number of parameters — that's _unfortunate_.

Project2015To2017/Transforms/PropertySimplificationTransformation.cs
andrew-boyarshin: If no merge issues occur and no duplicate field definition appears — we might indeed ignore this issue.

hvanbakel: Yeah I think this just looks weird because of how I merged it (I guess). I did a merge from master.

PR #177, opened 2018-08-23
Remove default Version properties

mungojam:
This removes any of the 3 possible Version properties if they are set to the default. Sometimes people set the version in CI or just never change it, so this leads to a simpler project file.

@andrew-boyarshin to get this new simplification to take advantage of your existing property simplicification I have moved the `PropertySimplificationTransformation` to later in the list of transforms. This seemed to have a knock-on effect on [this project](https://github.com/mungojam/Force.com-Toolkit-for-NET/blob/with-versions-cleared/src/ChatterToolkitForNET/ChatterToolkitForNET.csproj) which no longer has `<OutputPath>` in the unconditional property group, which it did have before [here](https://github.com/mungojam/Force.com-Toolkit-for-NET/blob/rc-fixes/src/ChatterToolkitForNET/ChatterToolkitForNET.csproj). Any idea which one is correct?

### DIFF
--- a/Project2015To2017/ProjectConverter.cs
+++ b/Project2015To2017/ProjectConverter.cs
@@ -177,7 +177,6 @@ namespace Project2015To2017
 			return new ITransformation[]
 			{
 				targetFrameworkTransformation,
-				new PropertySimplificationTransformation(),
 				new PropertyDeduplicationTransformation(),
 				new TestProjectPackageReferenceTransformation(this.logger),
 				new AssemblyReferenceTransformation(),
@@ -189,6 +188,7 @@ namespace Project2015To2017
 				new AssemblyAttributeTransformation(this.logger, this.conversionOptions.KeepAssemblyInfo),
 				new XamlPagesTransformation(this.logger),
 				new PrimaryUnconditionalPropertyTransformation(),
+				new PropertySimplificationTransformation(),
 				new EmptyGroupRemoveTransformation(),
 			};
 		}
--- a/Project2015To2017/Transforms/PropertySimplificationTransformation.cs
+++ b/Project2015To2017/Transforms/PropertySimplificationTransformation.cs
@@ -152,6 +152,9 @@ namespace Project2015To2017.Transforms
 					                         ):
 					case "RestorePackages" when ValidateDefaultValue("true"):
 					case "SchemaVersion" when ValidateDefaultValue("2.0"):
+					case "AssemblyVersion" when ValidateDefaultValue("1.0.0.0"):
+					case "FileVersion" when ValidateDefaultValue("1.0.0.0"):
+					case "Version" when ValidateDefaultValue("1.0.0"):
 					// Conditional platform default values
 					case "PlatformTarget" when parentConditionHasPlatform
 					                           && child.Value == parentConditionPlatform
--- a/Project2015To2017Tests/PropertySimplificationTransformationTest.cs
+++ b/Project2015To2017Tests/PropertySimplificationTransformationTest.cs
@@ -46,6 +46,9 @@ namespace Project2015To2017Tests
     </SccProvider>
     <NuGetPackageImportStamp>
     </NuGetPackageImportStamp>
+    <AssemblyVersion>1.0.0.0</AssemblyVersion>
+    <FileVersion>1.0.0.0</FileVersion>
+    <Version>1.0.0</Version>
   </PropertyGroup>
   <PropertyGroup Condition="" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' "">
     <DebugSymbols>true</DebugSymbols>


hvanbakel:
So removing this means that assembly version becomes the same as the version, that’s different from what it was. I would only remove that if the version is also 1.0.0.0 if it is not that might have been intentional and should not be changed.

As far as I know that’s the equivalent of setting 1.0.* which is no longer allowed if you want to keep your build deterministic.

Project2015To2017/ProjectConverter.cs
```diff
@@ -189,6 +188,7 @@ private IReadOnlyCollection<ITransformation> TransformationsToApply(bool modernP
 				new AssemblyAttributeTransformation(this.logger, this.conversionOptions.KeepAssemblyInfo),
 				new XamlPagesTransformation(this.logger),
 				new PrimaryUnconditionalPropertyTransformation(),
+				new PropertySimplificationTransformation(),
```
andrew-boyarshin: No, you can't move `PropertySimplificationTransformation` here.
* `PropertyDeduplicationTransformation` depends on simplification
* `PrimaryUnconditionalPropertyTransformation` depends on `PropertyDeduplicationTransformation`

So no way this can be moved this late. If you want to have properties, generated by `AssemblyAttributeTransformation` simplified — better move `AssemblyAttributeTransformation` to be before `PropertySimplificationTransformation`.

Project2015To2017/Transforms/PropertySimplificationTransformation.cs
```diff
@@ -152,6 +152,9 @@ public void Transform(Project definition)
 					                         ):
 					case "RestorePackages" when ValidateDefaultValue("true"):
 					case "SchemaVersion" when ValidateDefaultValue("2.0"):
+					case "AssemblyVersion" when ValidateDefaultValue("1.0.0.0"):
```
andrew-boyarshin: What if user specifies some version property to be `1.0`? Should we remove that too?

Project2015To2017/ProjectConverter.cs
mungojam: I remember now, the thing that meant I had to move it after the `PrimaryUnconditionalPropertyTransformation` was this bit in there:

```xml
			var propertyGroup = project.PrimaryPropertyGroup();
			propertyGroup.Add(project.AssemblyAttributeProperties);
```

Until that point, the `project.AssemblyAttributeProperties` haven't made it into the main property group, so they aren't available to be simplified. 

A few other options I thought of here:
1. Get rid of `AssemblyAttributeProperties` and just chuck them all in to PrimaryPropertyGroup. That would affect a lot of tests and might make users with custom transforms have to work harder (I think that includes me as I do custom things with the versions).
2. Make it a derived property that just fetches any of the associated properties from `project.PrimaryPropertyGroup` after they have been put there. I'd need to handle defaults here too though.
3. I could add them to the `project.PrimaryPropertyGroup` in the AssemblyAttributeTransform rather than the one above (or as well as and rely on the deduplication transform).

I think I will go for 3 as the simplest for now

andrew-boyarshin: Yeah, 3rd is the best one, but 2nd can in addition to 3rd be implemented not to break code relying on presence of `AssemblyAttributeProperties`.
But the second part is terrible, don't rely on deduplication for that — that transformation is for conditional `PropertyGroup`s only.

mungojam:
@hvanbakel 
> So removing this means that assembly version becomes the same as the version, that’s different from what it was. I would only remove that if the version is also 1.0.0.0 if it is not that might have been intentional and should not be changed.

The new SDK no longer copies versions between the different attributes if they are missing. That is why in the AssemblyAttributeTransform, I manually do the copy if the `AssemblyFileVersion` is missing. So I believe this is not a problem because it's already been dealt with there:

```csharp
//The AssemblyInfo behaviour was to fallback on the AssemblyVersion for the file version
//but in the new format, this doesn't happen so we explicitly copy the value across
CreateElementIfNotNullOrEmpty(assemblyAttributes.FileVersion, "FileVersion") ?? CreateElementIfNotNullOrEmpty(assemblyAttributes.Version, "FileVersion")
```

> As far as I know that’s the equivalent of setting 1.0.* which is no longer allowed if you want to keep your build deterministic.

1.0.* doesn't work at all in the new format whether it's in AssemblyInfo or csproj and throws an error on build. At the moment we just copy it across though we should also warn the user that they will need to come up with a solution. So I don't believe we are picking up any default behaviour by removing the version. It is what VS does when you create a new .net core project. It has nothing other than the TFM specified and it gets 1.0.0.0 for `AssemblyVersion` and `AssemblyFileVersion` and 1.0.0 for the `Version`

Project2015To2017/Transforms/PropertySimplificationTransformation.cs
mungojam: Fixing this now, I must have imagined doing it earlier

Project2015To2017/Transforms/PropertySimplificationTransformation.cs
```diff
@@ -212,6 +215,14 @@ bool ValidateConditionedDefaultValue(string @default)
 				{
 					return (valueLower == @default) && (!hasGlobalOverride || globalOverrideLower == @default);
 				}
+				
+				bool ValidateDefaultVersion()
+				{
+					return ValidateDefaultValue("1.0.0.0")
```
andrew-boyarshin: What if user specifies 1.0.0.0.0? I mean, who would be dumb enough, but should we handle that as well?
Like continuously `TrimEnd` `.0` and see if only `1` is left?

Project2015To2017/ProjectConverter.cs
mungojam: Ok, I've now just changed it to do it in `AssemblyAttributeTransform`. There might be scope to do something better in future to reduce the duplication but it's no worse than it was.

andrew-boyarshin:
And it would make the PR look more pleasant if you could squash your commits, since there were multiple attempts and revert commits usually are something to avoid. We could of course merge it with squash (since you have no GPG signature no content would be lost), but the wish to be fully the author of your work is common and understandable.

andrew-boyarshin:
And what's up with AppVeyor? Lots of test failures even on seemingly unrelated code.

mungojam:
> And it would make the PR look more pleasant if you could squash your commits, since there were multiple attempts and revert commits usually are something to avoid. We could of course merge it with squash (since you have no GPG signature no content would be lost), but the wish to be fully the author of your work is common and understandable.

There are different opinions depending on the project team. Some projects such as LineageOS make heavy use of reverts. A lot of teams would never edit a pushed branch because it can leave people with detached HEAD which might be tricky if they had also branched off to work on some other change (like I did earlier in the week).

Personally I prefer the PR to tell a story and then to squash when merging so people can see how decisions came about, but I know what you mean about the GPG etc. I might set that up in future, but for now I don't mind a github squash and merge which should leave my commits intact in this PR.

mungojam:
> And what's up with AppVeyor? Lots of test failures even on seemingly unrelated code.

These failures were all in the tests of the `AssemblyAttributeTransform` which are now fixed. I should have run them before committing.

mungojam:
An example of a closed PR with multiple commits from react: https://github.com/facebook/react/pull/13416

I know it's not as noisy as mine, but it looks like their principle is not to rebase within the PR

When merged, the PR then ends up as a nice single commit with a clear link to the PR and hovering over it shows a nice summary of the changes:
image

andrew-boyarshin:
@mungojam the difference between LineageOS (writing this from LineageOS-powered phone)  is that LOS iterates quickly and makes new releases every week. It is natural that bad things get reverted and reapplied later with fixes, especially given patch-oriented approach (oh, so Gerrit) and not branch-oriented one. This project however has a lot less contributors, far slower release cycle, thus allowing for neat commit history.
And yet I see what you mean about preserving the history, if that's the way you want things to be — only waiting for @hvanbakel approval and merge.

mungojam:
Yeah good point about the difference with lineageos and I do find their commit history looks messy.

We've just moved to git at work for 10s of projects, so I'm trying to decide on what workflows and rules to set there too. A tricky balance

Project2015To2017/Transforms/PropertySimplificationTransformation.cs
mungojam: I went for covering valid version numbers. 1.0.0.0.0 wouldn't have worked in the old world so I don't think we need to bother covering it whereas 1.0.0 would have worked and I think 1.0 and 1 would have too.

andrew-boyarshin: Ok, seems good. If user does stupid things there is no need to nurture them.

PR #180, opened 2018-08-24
Convert all file writes to output as UTF-8

laurencee:
A while back in PR #17 the files were written out as UTF-8 which is the default.

The file writing process was changed from a streamwriter with UTF-8 encoding over to File.WriteAllText but UTF-8 seems to have been dropped so this PR adds it back.


hvanbakel:
That seems like an omission, thanks for fixing that!

mungojam:
Thanks @laurencee I think I introduced that error. I've now learnt some new things!

PR #182, opened 2018-08-28
Rename transformations for consistency

andrew-boyarshin:
* Rename transformations for consistency
* Add `ILegacyOnlyProjectTransformation` and `IModernOnlyProjectTransformation` for future PR
* Rename transformation tests to match
* Move one test to its own `ImportsTargetsFilterPackageReferencesTransformationTest` file
* Remove unused using statements

### DIFF
--- a/Project2015To2017/ProjectConverter.cs
+++ b/Project2015To2017/ProjectConverter.cs
@@ -173,7 +173,7 @@ namespace Project2015To2017
 
 		private IReadOnlyCollection<ITransformation> TransformationsToApply(bool modernProject)
 		{
-			var targetFrameworkTransformation = new TargetFrameworkTransformation(
+			var targetFrameworkTransformation = new TargetFrameworkReplaceTransformation(
 				this.conversionOptions.TargetFrameworks,
 				this.conversionOptions.AppendTargetFrameworkToOutputPath);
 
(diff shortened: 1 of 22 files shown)


hvanbakel:
Can we do everything but the renames, that's technically breaking. Or maybe we just don't really care and create a 3.0.0 for the library only and keep the tool at 2.0.0

andrew-boyarshin:
@hvanbakel what can be done in minor versions is actually debatable. If no feature I removed, full functionality intact — can rename be counted breaking change? SharpDX for example does not count it as such (4.0->4.1 was highly breaking).

hvanbakel:
And on the other end of the spectrum is Newtonsoft json. I'll see if I can
merge this. I guess I'll just rev both as 3.0.0.

andrew-boyarshin:
Woah, 3.0 release already? I have had more changes pending for that kind of jump! I should've mentioned it in the description.

hvanbakel:
Yes, I thought about versioning yesterday for a while and for most users the version no longer matters because they use it as a global tool so they will run latest anyway.
For the few consumers (given that the global tool has more downloads already) of the nupkg, they can just decide if they want to upgrade. I’m not trying to take the versioning to the extreme but just keep it somewhat semantic.

PR #183, opened 2018-08-29
Fix #173

### LINKED ISSUE #173: Unsupported project types are crashing the tool
I take it the intended behaviour is that the tool will report an error but continue to run for supported project types. That seems to be the ideal since people may be running a batch conversion across many folders knowing that they might have some unsupported projects.

Anyway, when I run against the full Force.com example, using `Project2015To2017.Console.exe .\` at the root of the repository from commit: 7b0d1fc24864ad8026d93579779d19c95b9cdfe9 I get:

> info No nuspec found, skipping package configuration.
fail This project type is not supported for conversion.
info No nuspec found, skipping package configuration.
fail TargetFramework cannot be determined from project file. The scenario is not supported and highly bug-prone.

> Unhandled Exception: System.NotSupportedException: OutputType AppContainerExe is not supported.
   at Project2015To2017.Reading.ProjectPropertiesReader.ToApplicationType(String outputType)
   at Project2015To2017.Reading.ProjectPropertiesReader.Read(Project project)
   at Project2015To2017.Reading.ProjectReader.Read(FileInfo projectFile)
   at Project2015To2017.ProjectConverter.ProcessFile(FileInfo file, Solution solution, ProjectReference reference)
   at Project2015To2017.ProjectConverter.<Convert>d__5.MoveNext()
   at System.Linq.Enumerable.WhereEnumerableIterator`1.MoveNext()
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at Project2015To2017.Console.Program.ConvertProject(Options options)
   at CommandLine.ParserResultExtensions.WithParsed[T](ParserResult`1 result, Action`1 action)
   at Project2015To2017.Console.Program.Main(String[] args)

This happens with the released RC version and the new PR

The result is that nothing gets converted

hvanbakel:
Warn about the output type becoming a win exe but then just convert it. It will still work as a windows 8 store app according to [here](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/target-appcontainerexe-compiler-option#remarks).

### DIFF
--- a/Project2015To2017/Reading/ProjectPropertiesReader.cs
+++ b/Project2015To2017/Reading/ProjectPropertiesReader.cs
@@ -169,7 +169,7 @@ namespace Project2015To2017.Reading
 			throw new NotSupportedException($"Target framework {targetFramework} is not supported.");
 		}
 
-		private static ApplicationType ToApplicationType(string outputType)
+		private ApplicationType ToApplicationType(string outputType)
 		{
 			if (string.IsNullOrWhiteSpace(outputType))
 			{
@@ -178,9 +178,15 @@ namespace Project2015To2017.Reading
 
 			switch (outputType.ToLowerInvariant())
 			{
-				case "exe": return ApplicationType.ConsoleApplication;
-				case "library": return ApplicationType.ClassLibrary;
-				case "winexe": return ApplicationType.WindowsApplication;
+				case "exe":
+					return ApplicationType.ConsoleApplication;
+				case "library":
+					return ApplicationType.ClassLibrary;
+				case "appcontainerexe":
+					this._logger.LogWarning("Converting AppContainerExe to more generic Windows exe.");
+					return ApplicationType.WindowsApplication;
+				case "winexe":
+					return ApplicationType.WindowsApplication;
 				default: throw new NotSupportedException($"OutputType {outputType} is not supported.");
 			}
 		}


andrew-boyarshin:
Why did you not introduce new `ApplicationType` enumeration value? That seems more simple and will preserve the original value.

hvanbakel:
And then on the writing side collapse them? I can do that, as far as I could tell this option is no longer supported so there’s nothing really that we an do about that. I see your point of not losing that information up until the point where we write, let me update that.

andrew-boyarshin:
@hvanbakel wait, it's not supported? I mean, in MSBuild `.props` and `.targets` sources I've seen things related to that.

hvanbakel:
I wasn’t able to find the proper output type for that. If it is supported than the entire PR is useless.

andrew-boyarshin:
It should be. Example:
```xml
  <PropertyGroup>
    <TargetExt Condition="'$(TargetExt)' == '' and '$(OutputType)'=='exe'">.exe</TargetExt>
    <TargetExt Condition="'$(TargetExt)' == '' and '$(OutputType)'=='winexe'">.exe</TargetExt>
    <TargetExt Condition="'$(TargetExt)' == '' and '$(OutputType)'=='appcontainerexe'">.exe</TargetExt>
    <TargetExt Condition="'$(TargetExt)' == '' and '$(OutputType)'=='library'">.dll</TargetExt>
    <TargetExt Condition="'$(TargetExt)' == '' and '$(OutputType)'=='module'">.netmodule</TargetExt>
    <TargetExt Condition="'$(TargetExt)' == '' and '$(OutputType)'=='winmdobj'">.winmdobj</TargetExt>
  </PropertyGroup>
```
And there are more ocurrences of `appcontainerexe` in the same file. It certainly is a feature of Roslyn (hence the documentation page you linked).

andrew-boyarshin:
The XML extract from `Microsoft.Common.CurrentVersion.targets` is also a good source of supported `$(OutputType)` values.

hvanbakel:
Nice I'll revisit

PR #184, opened 2018-08-29
Refactor transformations logic

andrew-boyarshin:
* Add `ITransformationSet` interface and noop implementation
* Add `BasicReadTransformationSet` to be used as default (enough for analyzers)
* Add `ChainTransformationSet` that allows chaining multiple sets
* Add `Vs15TransformationSet` that contains all implemented transformations
* Add `TargetTransformationExecutionMoment` enum (Normal/Early/Late)
* Add `ITransformationWithDependencies` and `ITransformationWithTargetMoment`
* Refactor `ProjectConverter.TransformationsToApply()` to do topological sort (ordering) on a transformation graph (to be precise - reverse graph) by stages (`TargetTransformationExecutionMoment`)
* Fix `NuGetPackageTransformation` to have NuGet casing (I certainly remember it being fixed some time ago)


Project2015To2017/Transforms/Vs15TransformationSet.cs
```diff
@@ -0,0 +1,43 @@
...
+				new PrimaryProjectPropertiesUpdateTransformation(),
+				new EmptyGroupRemoveTransformation(),
+				// VS15 migration
+				new TestProjectPackageReferenceTransformation(logger),
+				new AssemblyFilterPackageReferencesTransformation(),
+				new AssemblyFilterHintedPackageReferencesTransformation(),
+				new AssemblyFilterDefaultTransformation(),
+				new ImportsTargetsFilterPackageReferencesTransformation(),
+				new FileTransformation(logger),
+				new XamlPagesTransformation(logger),
```
mungojam: Looks like this is missing `AssemblyAttributeTransformation` but maybe I have misunderstood how this is all working. I'm finding it much harder to follow what's going on

mungojam: I see now that the two are combined above in `Instance`. So then what does `TrueInstance` mean?

Project2015To2017/ProjectConverter.cs
```diff
@@ -172,37 +188,114 @@ internal static bool Validate(FileInfo file, ILogger logger)
...
+			var res = new LinkedList<ITransformation>();
+			var mappings = new Dictionary<string, int>();
+			for (var i = 0; i < count; i++)
+			{
+				var transformation = source[i];
+				if (transformation == null)
+				{
+					throw new ArgumentNullException(nameof(transformation),
+						"Transformation set must not contain null items");
+				}
```
mungojam: I feel that you favour performance over readability too much for the needs of this project. To me, this check and addition of the items to the list would be much clearer as an `Any()` call and a `Select().ToDictionary()` call. In practise nobody is trying to convert 10,000 projects and even with that many the difference would probably be minimal.

andrew-boyarshin: I do use Linq methods, but in cases where I believe things will get simplified. This `for` loop is natural, its designation is obvious from looking at the code. If I replace that with 2 lines of Linq garbage — from my point of view it will get more complicated to grasp the meaning (I "love" `ToDictionary" arguments — who could've imagined less obvious API?). It's not performance optimization, rather it is just as I've been step-by-step writing this code. When I've been reviewing this code myself I didn't feel the urge to replace things with Linq.

Project2015To2017/ProjectConverter.cs
```diff
@@ -24,11 +24,16 @@ public sealed class ProjectConverter
 		private readonly ILogger logger;
 		private readonly ConversionOptions conversionOptions;
 		private readonly ProjectReader projectReader;
+		private readonly ITransformationSet transformationSet;
 
-		public ProjectConverter(ILogger logger, ConversionOptions conversionOptions = null)
+		public ProjectConverter(
+			ILogger logger,
+			ConversionOptions conversionOptions = null,
+			ITransformationSet transformationSet = null)
```
hvanbakel: Is this parameter really optional? I would just have the implementor specify one and not default to anything but throw instead.

andrew-boyarshin: That was intended to be required argument, but later I changed it to optional to preserve API contract. And it is used in analyzers (after conversion).

Project2015To2017/ProjectConverter.cs
```diff
@@ -148,8 +154,18 @@ private Project ProcessFile(FileInfo file, Solution solution, ProjectReference r
 				transform.Transform(project);
 			}
 
-			foreach (var transform in TransformationsToApply(project.IsModernProject))
+			foreach (var transform in TransformationsToApply())
 			{
+				if (project.IsModernProject && transform is ILegacyOnlyProjectTransformation)
```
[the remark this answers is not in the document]
andrew-boyarshin: That's simple actually. @mungojam mentioned the need to run new/modified transformations on already converted projects. If we move this logic to transformations (I've done that at first and then realized all that) there will be no way to force execution of specific transform (except to modify the code/add new parameters to Transform interface method). Now the only thing that should be done is add necessary _force_ logic to ProjectConverter.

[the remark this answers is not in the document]
hvanbakel: If you want to rerun then you should make sure it is deterministic, right? 

I don’t see how you ever want to force anything, we should strive for deterministic transforms. If a file no longer exists (like nuspec) just no-op, that’s fine.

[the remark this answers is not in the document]
andrew-boyarshin: Yeah, but deterministic is not something guaranteed to be preserved between versions (behavior of FileTransformation changed greatly in my earlier PR). User wants to take advantage of improved conversion behavior, so they pass some CLI argument and force execution despite the fact that it is already converted (`IsModernProject` is `true`).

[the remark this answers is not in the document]
hvanbakel: I can see all that, but I’m missing the point as to why this check can’t be done inside the transformation itself. I see what you’re trying to say about a force option, but I don’t see yet why you would need such a thing. 

I think we should assume that the 99% of users will use it as a global tool and just run it without any parameters except for a project/solution. I don’t see a lot of users sifting through the docs in order to figure out how they can do incremental runs. The 2015 -> 2017 change was a big one, it will probably be a very long time before a similar change (needing tooling) will occur.

andrew-boyarshin:
I've addressed ProjectConverter constructor feedback and pushed amended commit. Other notes are mostly on overengineering (although I believe the code that does things in O(n) instead of O(2n) and still looks pretty readable (6 codelines instead of 2 long lines) should not cause such feedback).

hvanbakel:
Don't take it the wrong way, that's why I approved the changes. The things I brought up are not structural issues, I always feel like the discussions are very informative as to why certain things were solved in a specific way. Everyone is biased by their normal code practices which makes opensource refreshing

andrew-boyarshin:
I'm not taking it the wrong way, but I do wonder why people think Linq is the ultimate answer for everything (not to mention actual Linq that is statements starting with `from`, limited as ever). Trivial things (filter and call method on found items) can indeed be simplified with Linq, that's natural. But when people try to express more complex logic in Linq they tend to omit the additional code readability drop and performance overhead they introduce by doing that. Linq doesn't simply solve the issue, it provides an abstraction level for any `IEnumerable<T>`, but developer must still remember that the inner costs are hidden, not removed. In my projects I have ReSharper suggest a number of cases that can be simplified, including replacement with Linq. But when I apply the quick-fix, I understand that I can no longer perceive what this code is for. Furthermore, with Linq it gets increasingly harder to introduce changes, since you can't deviate from the pattern Linq enforces.
And yeah, it's hard to disagree that everyone is biased according to their own set of experiences.

mungojam:
If the set of transformations no longer has a default that will make it a bit trickier to use the nuget package. It's probably fine if we can add something to the readme showing how to use the package for a default run and for one with an additional custom transform

mungojam:
The linq one is definitely style and preference. If you ever get into F#, you might see where I'm coming from. I don't use it everywhere but default to it until I need to improve performance. 

I generally find it much easier to establish if code is right for example I have a gaurantee that a Select will give me a 1 to 1 mapping and an Any won't be editing anything but just checking each item.

It is a different dialect, familiar to those with functional programming experience as they are all standard opeationsm. Over time the performance differences will be hopefully removed by better compilers and JITer

andrew-boyarshin:
@mungojam yeah, I know that was a good attempt at bringing functional-style into C#, and I like some elements of that. I've written this code in imperative fashion. There are pros and cons to that. Still, we can't write code that satisfies everyone. If there are no further notes or major issues I believe this can be merged.
P.S. I am working on next PR now, related to CLI.

hvanbakel:
Nice @andrew-boyarshin, there’s a reason I made both of you collaborators, I don’t always have the time to merge/review quickly enough. 

The discussion on using linq or not is mostly opnions I guess. However, I would keep this in mind
> "Programs should be written for people to read, and only incidentally for machines to execute." -- Structure and Interpretation of Computer Programs

Linq does make code more readable and virtually everyone who writes C# knows how to read it. There’s absolutely overhead in there but that would be something to address in hotspots in your code after detailed analysis of where those hotspots are. Especially on big(ger) teams this becomes paramount because you can more easily grow the team to work on the non performance critical parts.

And to be completely clear, I wouldn’t be the one writing the performant code, I would need to google what stackalloc does but that’s fine (and you haven’t used that yet).

PR #186, opened 2018-08-30
Refactor CLI

### LINKED ISSUE #163: Running with no parameters outputs '### Performing 2nd pass...'
A couple of things here:

> C:\Users\mark\source\repos\Force.com-Toolkit-for-NET>csproj-to-2017

1. It outputs the message when it should just output the usage instructions or an error or something
2. It only sometimes outputs it, which makes me wonder if we need an extra flush or something

andrew-boyarshin:
* Separate worker logic into CommandLogic
* Separate worker logic into Evaluation, Migration and Analysis paths
* Replace CommandLine with Microsoft.DotNet.Cli.CommandLine
* Replace hacky ConsoleLogger with Serilog
* Fix all known CLI drawbacks

Fix #163


andrew-boyarshin:
Some troubles with `RestoreAdditionalSources` on AppVeyor, investigating...

Project2015To2017.Console/Project2015To2017.Console.csproj
```diff
@@ -15,22 +14,29 @@
...
+		</RestoreAdditionalProjectSources>
 	</PropertyGroup>
 
 	<PropertyGroup Condition="$(Pack) != 'true'">
 		<TargetFrameworks>net461;netcoreapp2.1</TargetFrameworks>
 	</PropertyGroup>
 
 	<PropertyGroup Condition="$(Pack) == 'true'">
-		<ToolCommandName>csproj-to-2017</ToolCommandName>
+		<ToolCommandName>dotnet-migrate-2017</ToolCommandName>
```
hvanbakel: I think we should be doing this in a new separate console project, I don’t want to kill the current CLI command because people might be upgrading and that’s very unexpected.

Project2015To2017.Console/ConsoleLogger.cs
```diff
@@ -1,302 +0,0 @@
...
-using Microsoft.Extensions.Logging.Console;
-using Microsoft.Extensions.Logging.Console.Internal;
-using System;
-using System.Runtime.InteropServices;
-using System.Text;
-
-namespace Project2015To2017.Console
-{
-
-	public class ConsoleLogger : ILogger
```
hvanbakel: I actually like that we’re getting rid of this one

Project2015To2017.Console/Project2015To2017.Console.csproj
andrew-boyarshin: And keep it with buggy CLI? If we define this to be 4.0.0 (I expected 3.0.0 - major version increase after this and a couple more PRs, not so early), we can introduce this breaking change and keep it this way (drop old CLI), but specify it in release notes (I mean, write real change notes that specify changes in behaviors and such).

andrew-boyarshin: I mean, no problem, easily done, but should we really keep old CLI?

andrew-boyarshin:
Done. @hvanbakel @mungojam ?

Project2015To2017.Console/Program.cs
```diff
@@ -23,6 +23,10 @@ private static void ConvertProject(Options options)
 			var convertedProjects = new List<Project>();
 
 			ILogger logger = new ConsoleLogger("console", (s, l) => l >= LogLevel.Information, true);
+
+			logger.LogWarning("csproj-to-2017 is deprecated and will be removed soon");
+			logger.LogInformation("Consider migrating to Project2015To2017.Migrate2017.Tool (dotnet migrate-2017)");
```
hvanbakel: Nice

hvanbakel:
The reason appveyor fails is that it doesn't know about your custom nuget source. I would update appveyor.yml on line 11 to have `nuget restore --source https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json --source https://api.nuget.org/v3/index.json`

andrew-boyarshin:
@hvanbakel but it did work before. After I replaced `RestoreAdditionalSources` with `RestoreAdditionalProjectSources` in my 1st comment above it ran the full build successfully. Why not recognize the source now?

andrew-boyarshin:
And why not `dotnet restore` instead? That seems more modern.

hvanbakel:
I think at the time when this was created dotnet wasn't there yet.

As far as I know currently doing just dotnet build is sufficient, restore
is implied now. We could try that instead maybe that picks up the source.

andrew-boyarshin:
Yeah, but `dotnet restore` is still present for explicit package restore (e.g. for in CI environments). I'll try that.

andrew-boyarshin:
It worked! `dotnet restore` is the way to go.

hvanbakel:
I think this is good to go. The appveyor file is not entirely correct though (you swapped the name of the tools but didn't touch the keys. I can update that later.

PR #187, opened 2018-08-31
Fix for #173

hvanbakel:


hvanbakel:
For some weird reason I got a crash when building before. Maybe that was a typo... I don't know but it does work now.

andrew-boyarshin:
I already have (while GitHub is down). :)

Issue #176, opened 2018-08-23
Unix-like OS support


andrew-boyarshin:
Thanks to the latest work on Unix-like OS support, we now have working SolutionReader and ProjectReference reader. I am afraid that's not the end.

```
warn AssemblyInfo file '/home/andrew/dev/NAudio/NAudio/Properties\AssemblyInfo.cs' not found
```
Well, you see the issue. File exists but not found. We need to make sure to call slash conversion routine everywhere we work with paths.

hvanbakel:
@andrew-boyarshin what version did you get that with? As the code in question (assuming you don't have a full path in your include) is this `Path.GetFullPath(Path.Combine(projectPath, x))` which as far as I know should throw an `ArgumentException` on illegal chars. Is `\` not considered an illegal char for a path on ubuntu?

andrew-boyarshin:
@hvanbakel 2.0.0 release as global tool, latest .NET Core SDK (2.1.401)+runtime, Ubuntu 18.04.

hvanbakel:
So when you do `Path.FullPath(Path.Combine(“/home/andrew”, “Properties\\assemblyinfo.cs”))` that does not throw (it does on windows for the forward slash).

andrew-boyarshin:
@hvanbakel will run some checks tomorrow. I'm in final preparation stage for new PR (related to transformations loading and ordering).

hvanbakel:
Ok, if you have issues that you’re working on, could you assign those to yourself, I want to make sure we’re not doing the same thing.

andrew-boyarshin:
@hvanbakel what am I doing wrong?
```
Microsoft (R) Visual C# Interactive Compiler version 2.10.0.63119
Copyright (C) Microsoft Corporation. All rights reserved.

Type "#help" for more information.
> Path.GetFullPath(Path.Combine("/home/andrew", "Properties\\assemblyinfo.cs"));
> Console.WriteLine(Path.GetFullPath(Path.Combine("/home/andrew", "Properties\\assemblyinfo.cs")));
C:\home\andrew\Properties\assemblyinfo.cs
```

andrew-boyarshin:
And yeah, that was `csi` on Win10:
```
Visual Studio 2017 Developer Command Prompt v15.9.0-pre.1.0
```

andrew-boyarshin:
And I don't see it stated explicitly in documentation why it should throw any exception.

hvanbakel:
[...]

PR #188, opened 2018-08-31
Improve CLI, fix bugs

andrew-boyarshin:
* Make specialized `ProjectConverter` methods public
* Rewrite projects selection logic from scratch for modern CLI tool
* Add globbing as a fallback to modern CLI tool
* Reduce default log verbosity (mark common messages as _Debug_ instead of _Informational_)
* Add Trace level logging to modern CLI tool
* Make evaluate command use only _W00x_ and _W01x_ analyzer diagnostic sets
* Print unsupported solution projects in evaluate command
* Add `--force-transformations` CLI option
* Make `ConditionEvaluator` much safer against unknown conditionals
* Revert `<SolutionDir />` removal from `PropertySimplificationTransformation` (it is not the default in CPS, used when running MSBuild on single project instead of solution). Refer to #165.
* Fix NuGet name case in log messages
* Use `FileInfo` as argument in `SolutionReader`
* Fix missing `<GenerateAssemblyInfo>false</GenerateAssemblyInfo>` after #177.
* Save our private field naming convention for JetBrains products

[Good attempt at creating normal changelog](https://github.com/hvanbakel/CsprojToVs2017/wiki/Changelog)

### DIFF
--- a/Project2015To2017.Console/Program.cs
+++ b/Project2015To2017.Console/Program.cs
@@ -1,10 +1,13 @@
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
 using CommandLine;
 using Microsoft.Extensions.Logging;
 using Project2015To2017.Analysis;
 using Project2015To2017.Definition;
+using Project2015To2017.Reading;
 using Project2015To2017.Transforms;
-using System.Collections.Generic;
-using System.Linq;
 
 namespace Project2015To2017.Console
 {
@@ -30,7 +33,7 @@ namespace Project2015To2017.Console
 			foreach (var file in options.Files)
 			{
 				var projects = new ProjectConverter(logger, Vs15TransformationSet.Instance, conversionOptions)
-					.Convert(file)
+					.Convert(file, logger)
 					.Where(x => x != null)
 					.ToList();
 				convertedProjects.AddRange(projects);
@@ -68,7 +71,7 @@ namespace Project2015To2017.Console
 			foreach (var file in options.Files)
 			{
 				var projects = new ProjectConverter(logger, BasicReadTransformationSet.Instance, conversionOptions)
-					.Convert(file)
+					.Convert(file, logger)
 					.Where(x => x != null)
 					.ToList();
 				convertedProjects.AddRange(projects);
@@ -79,6 +82,81 @@ namespace Project2015To2017.Console
 			foreach (var project in convertedProjects)
 			{
 				analyzer.Analyze(project);
+			}
+		}
+
+		private static IEnumerable<Project> Convert(this ProjectConverter self, string target, ILogger logger)
+		{
+			var extension = Path.GetExtension(target) ?? throw new ArgumentNullException(nameof(target));
+			if (extension.Length > 0)
+			{
+				var file = new FileInfo(target);
+
+				switch (extension)
+				{
+					case ".sln":
+					{
+						var solution = SolutionReader.Instance.Read(file, logger);
+						foreach (var project in self.ProcessSolutionFile(solution))
+						{
+							yield return project;
+						}
+						break;
+					}
+					case string s when ProjectConverter.ProjectFileMappings.ContainsKey(extension):
+					{
+						yield return self.ProcessProjectFile(file, null);
+						break;
+					}
+					default:
+						logger.LogCritical("Please specify a project or solution file.");
+						break;
+				}
+
+				yield break;
+			}
+
+			// Process the only solution in given directory
+			var solutionFiles = Directory.EnumerateFiles(target, "*.sln", SearchOption.TopDirectoryOnly).ToArray();
+			if (solutionFiles.Length == 1)
+			{
+				var solution = SolutionReader.Instance.Read(solutionFiles[0], logger);
+				foreach (var project in self.ProcessSolutionFile(solution))
+				{
+					yield return project;
+				}
+
+				yield break;
+			}
+
+			var projectsProcessed = 0;
+			// Process all csprojs found in given directory
+			foreach (var fileExtension in ProjectConverter.ProjectFileMappings.Keys)
+			{
+				var projectFiles = Directory.EnumerateFiles(target, "*" + fileExtension, SearchOption.AllDirectories).ToArray();
+				if (projectFiles.Length == 0)
+				{
+					continue;
+				}
+
+				if (projectFiles.Length > 1)
+				{
+					logger.LogInformation($"Multiple project files found under directory {target}:");
+				}
+
+				logger.LogInformation(string.Join(Environment.NewLine, projectFiles));
+
+				foreach (var projectFile in projectFiles)
+				{
+					// todo: rewrite both directory enumerations to use FileInfo instead of raw strings
+					yield return self.ProcessProjectFile(new FileInfo(projectFile), null);
+					projectsProcessed++;
+				}
+			}
+
+			if (projectsProcessed == 0)
+			{
+				logger.LogCritical("Please specify a project file.");
 			}
 		}
 	}
--- a/Project2015To2017.Migrate2017.Tool/CommandLogic.cs
+++ b/Project2015To2017.Migrate2017.Tool/CommandLogic.cs
@@ -1,48 +1,146 @@
-using System;
-using System.Collections.Generic;
-using System.Linq;
+using DotNet.Globbing;
 using Project2015To2017.Analysis;
 using Project2015To2017.Definition;
+using Project2015To2017.Reading;
 using Project2015To2017.Transforms;
 using Project2015To2017.Writing;
+using Serilog;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.IO;
+using System.Linq;
 
 namespace Project2015To2017.Migrate2017.Tool
 {
 	public class CommandLogic
 	{
 		private readonly Microsoft.Extensions.Logging.ILogger genericLogger;
+		private ImmutableArray<(string path, string extension)> files;
+		private readonly ImmutableArray<string> extensions;
 
 		public CommandLogic()
 		{
 			this.genericLogger = new Serilog.Extensions.Logging.SerilogLoggerProvider().CreateLogger(nameof(Serilog));
+			extensions = ProjectConverter.ProjectFileMappings.Keys.Concat(new[] {".sln"}).ToImmutableArray();
 		}
 
-		public IReadOnlyCollection<Project> ParseProjects(
+		public void DoProcessableFileSearch(bool force = false)
+		{
+			if (files != null && files.Length > 0 && !force)
+			{
+				Log.Verbose("Glob file list reevaluation skipped");
+				return;
+			}
+
+			Log.Verbose("Glob file list reevaluation started");
+			files = Directory.EnumerateFiles(Directory.GetCurrentDirectory(), "*.*", SearchOption.AllDirectories)
+				.Select(x => (path: x, extension: Path.GetExtension(x)?.ToLowerInvariant()))
+				.Where(x => !string.IsNullOrEmpty(x.extension))
+				.Where(x => extensions.Contains(x.extension))
+				.ToImmutableArray();
+			Log.Verbose("Glob file list reevaluation finished: {Count} items", files.Length);
+		}
+
+		public (IReadOnlyCollection<Project> projects, IReadOnlyCollection<Solution> solutions) ParseProjects(
 			IEnumerable<string> items,
 			ITransformationSet transformationSet,
 			ConversionOptions conversionOptions)
 		{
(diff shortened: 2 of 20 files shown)


Project2015To2017.Migrate2017.Tool/Program.cs
```diff
@@ -145,6 +149,12 @@ private static int ProcessArgs(ParseResult result)
 			OneOrMoreArguments()
 				.With("Target frameworks to be used instead of the ones in source projects", "frameworks"));
 
+		private static Option ForceTransformationsOption => Option(
+			"-f|--force-transformations",
+			"Force execution of transformations despite project conversion state by their specified names.",
```
mungojam: I'm not clear on what this message means, particularly `by their specified names`

Project2015To2017/ProjectConverter.cs
```diff
@@ -37,26 +37,34 @@ public sealed class ProjectConverter
 			this.projectReader = new ProjectReader(logger, this.conversionOptions);
 		}
 
+		[Obsolete("This method is deprecated, call Process[Project|Solution]File directly instead")]
```
mungojam: Is there an easy way I can use the globbing or folder-wide functionality when using this as a package? Before I could pass in a folder to this for it to process all our solutions which is handy in our setup

Project2015To2017/Reading/NuSpecReader.cs
```diff
@@ -24,14 +24,17 @@ public PackageConfiguration Read(FileInfo projectFile)
 
 			if (nuspecFiles.Length == 0)
 			{
-				this.logger.LogInformation("No nuspec found, skipping package configuration.");
+				this.logger.LogDebug("No nuspec found, skipping package configuration.");
```
mungojam: Nice! this should help to reduce the noise a lot

Project2015To2017/Reading/NuSpecReader.cs
```diff
@@ -24,14 +24,17 @@ public PackageConfiguration Read(FileInfo projectFile)
...
 				return null;
 			}
 
 			if (nuspecFiles.Length > 1)
 			{
-				this.logger.LogError($@"Could not read from nuspec, multiple nuspecs found: 
-{string.Join(Environment.NewLine, nuspecFiles.Select(x => x.FullName))}.");
+				this.logger.LogError(@"Could not read from nuspec, multiple nuspecs found.");
```
mungojam: I think you've inadvertently changed `:` to `.` before the list

Project2015To2017.Migrate2017.Tool/Program.cs
andrew-boyarshin: By transformations' specified names (like `"FileTransformation"`). Any ideas how to make this more clear?

Project2015To2017/Reading/NuSpecReader.cs
mungojam: I can see the difficulty though, since you are making the others ones for information only

Project2015To2017/ProjectConverter.cs
andrew-boyarshin: That was intended to be part of user-oriented Migrate2017.Tool functionality. If you control callsite (use this as a NuGet), you can handle things much better on your side rather than rely on library's built-in solutions. Library should provide solid and generic API, using which anyone can build their custom logic.

Project2015To2017/Reading/NuSpecReader.cs
andrew-boyarshin: No, that was fully intentional. Since following list are printed on `Informational` level, they might be not shown at all. The `:` would look weird and force user to think if they have a bug on their side.

Project2015To2017/ProjectConverter.cs
mungojam: I only want a tiny bit of customisation though (the custom transforms before and after). That's why I added it to this project. It would be annoying to have to roll my own directory searching when the code is all there in the project. Remember this conversion is a one time activity so I don't want to spend ages writing the code to do it.

andrew-boyarshin: I mean, is this that difficult (the following is written by memory without IDE)?
```cs
var converter = new ProjectConverter(...);
var projects = new List<Project>();
foreach (var x in Directory.EnumerateFiles(Directory.GetCurrentDirectory(), "*.csproj", SearchOption.AllDirectories))
    projects.Add(converter.ProcessProjectFile(new FileInfo(x), null));
```

andrew-boyarshin: But this piece of code will certainly do only what _you_ need, it will not try doing other things that are not needed in your scenario.

mungojam: Remember I am one of the few users of the package and you are taking a feature away from me while simultaneously implementing it in a bit of the code that I can't make easy use of. I don't work with files and folders as much as you might so no I couldn't have written the above from memory and I would probably get it wrong a couple of times before I got it right.

If we could just add `ProcessFolder`, with a globbing option (maybe even defaulted to `**/*.sln`, then I'll be happy

andrew-boyarshin: But that would force library to depend on additional NuGet package. If @hvanbakel is fine with that - why not. But that would tempt to use that when better solutions are present (like the one above). Globbing is really expensive, in CLI I first fetch all possible matching files (once in a process lifetime) and only then apply glob matching.

hvanbakel: I don't think there's an issue with having methods in the core library that we don't use in the cli. They are 2 different interfaces to the same functionality, they don't have to be the same.

I'll do a full review later today

Project2015To2017.Console/Program.cs
```diff
@@ -81,5 +84,80 @@ private static void ConvertProject(Options options)
 				analyzer.Analyze(project);
 			}
 		}
+
+		private static IEnumerable<Project> Convert(this ProjectConverter self, string target, ILogger logger)
```
hvanbakel: I wouldn't move this logic here. I do agree with you that this is a console concern, but this limits the usability of the nupkg for people extending it. 

While the vast majority of users will go through the CLI, I do like the ability to extend the core project. I can see a lot of small extensions being written to fix company specific things. I wouldn't want everyone forking the project for that.

I think we can leave this in the core library for people wrapping the nupkg and if we want to do something different for the console than we can write our own here.

Project2015To2017.Migrate2017.Tool/CommandLogic.cs
```diff
@@ -1,89 +1,198 @@
...
+		private readonly ImmutableArray<string> extensions;
 
 		public CommandLogic()
 		{
 			this.genericLogger = new Serilog.Extensions.Logging.SerilogLoggerProvider().CreateLogger(nameof(Serilog));
+			extensions = ProjectConverter.ProjectFileMappings.Keys.Concat(new[] {".sln"}).ToImmutableArray();
 		}
 
-		public IReadOnlyCollection<Project> ParseProjects(
+		public void DoProcessableFileSearch(bool force = false)
```
hvanbakel: Why does this use a field, this could/should be side effect free, right? This could be private?

Project2015To2017.Migrate2017.Tool/CommandLogic.cs
```diff
@@ -1,89 +1,198 @@
...
+			foreach (var pattern in items)
 			{
-				var projects = new ProjectConverter(genericLogger, transformationSet, conversionOptions)
-					.Convert(file)
-					.Where(x => x != null)
-					.ToList();
-				convertedProjects.AddRange(projects);
+				if (File.Exists(pattern))
+				{
+					var file = new FileInfo(pattern);
```
hvanbakel: We should handle/rewrite the slashes here so we don't introduce new cases of #176

Project2015To2017.Migrate2017.Tool/Program.cs
hvanbakel: This means very precise knowledge of the internals of the tool. Do you see a real use case for this? I don't expect users of a command line tool to know which transformations are available unless we put that in the help file as well.

Project2015To2017/ProjectConverter.cs
hvanbakel: I would just leave the `Obsolete` out it doesn't matter that we don't use it, it is useful for the nupkg consumers.

Project2015To2017.Migrate2017.Tool/CommandLogic.cs
hvanbakel: Maybe I'm missing the point of what you're trying to do here. So you call this once to set a field, but I can never get hold of the value of that field. So how am I supposed to use this.

If it would return the value I could pass that in somewhere else and this would be the one place where you hit the filesystem.

andrew-boyarshin:
Regarding "force" feature. Imagine user has converted their projects back in 1.4.0. Now is a good time to apply something new, e.g. `PropertySimplification` and `PropertyDeduplication`. With force switch user can override check and enforce execution even on already converted projects. Is it not a good reason to have a new feature, even if it is not a commonly used one?

mungojam:
Force is a useful feature for my use but only if it works via the nuget
package

Project2015To2017.Migrate2017.Tool/CommandLogic.cs
hvanbakel: So I would pass that true to reevaluate. That means I am trying to handle the case where, while the tool is running, new solutions/projects were added under the current directory?

hvanbakel:
First of all, I really appreciate all the hard work you put in to this but I think we have a different view of what the tool should and should not try to do.

The problem we're trying to solve is the lack of a migration path for msbuild files between the old and new schema, essentially handling the breaking change that Microsoft made in MSBuild. This tool is supposed to help users do that, by automating as much of the process as it possibly can as well as pointing them to future optimizations where possible (analyzers are a great addition). But, as soon as conversion happened successfully they're on their way and they can make the small incremental changes they've been doing in the old setup too.

Therefor, I seriously doubt if there is anyone that will run this tool multiple times, because:
1. At that point the diff between old project file and new project file is (probably) so small that you could've done it by hand. 
2. Rerunning means that the user 'knows' that there is a reason to rerun which implies that you could just as well do it by hand.
3. The 'need' to run isn't there anymore, I doubt anyone will just run a random tool and trust it to make sensible upgrades. The initial problem is solved, they're on the new build definition now.
This is where my comments are coming from, I think we're over complicating the tool for an edge case. The force feature just made that obvious to me. 
I would've loved to have statistics on this but I don't have those but from looking at the traffic to this github repo and where it is coming from does confirm this theory. As my single point of reference, I've run this tool on a vast number of projects and never specified a single argument other than a path.

I think that the best tools out there are the tools that solve a single problem and solve it well. More importantly they know what not to solve. I started this project because I believed/knew there was an issue and some people at Microsoft had said that while they knew it existed and it was on their backlog, it wasn't a priority for them. 

Looking at the open issues we have, those are all incremental upgrades that work well for the 'one time use' scenario. If you ran a previous version of the tool that did 90% of what the current tool does, too bad, but apparently it was sufficient for you otherwise you would've thrown away the result and did it by hand.

mungojam:
Interesting points yeah. I do agree with that being the focus of the tool and making sure that improvements are focused on that 90%  case. 

Maybe that's why I favoured the eager mode that would do a more opinionated first time conversion. Like you say, if it breaks anything, then I can manually fix it anyway, the new format makes that much easier. I like our project files to be as slimline as possible so that those in my team feel that they understand every line that's in it and feel confident working with it when needed.

I think I would use a force mode on my already converted projects if it just worked without me needing to do anything else. But I've only converted about 50% so far so maybe I'll never get around to it. It certainly isn't high priority for me to do.

Is one consideration how much time we actually have to review changes that are for a 1% case. I'm finding myself now reviewing things just to make sure that I don't lose a feature that I depend on

andrew-boyarshin:
I will address PR feedback in ~8 hours.

Project2015To2017.Migrate2017.Tool/CommandLogic.cs
andrew-boyarshin: One use can be that, yeah. But it might also be beneficial for possible future extension (or use of CLI tool as library, why the hell not?)

andrew-boyarshin:
@hvanbakel @mungojam reverted Convert obsoletion and feature set change. Is it good to go?

Project2015To2017.Migrate2017.Tool/Program.cs
hvanbakel: I think we can merge this in, I still think it's a bad idea to make type names part of your CLI interface. But I'm guessing that very few people (if any) will use it so it might never result in issues being logged.

hvanbakel:
I'll setup publishing for the package later today.

PR #193, opened 2018-09-08
Attempt at fixing slashes

hvanbakel:
Attempted fix for #176. I've left places where we compare a path by doing `StartsWith(x)` untouched for now. I'm not entirely sure if I should be transforming both sides there though, it's either both or none.


andrew-boyarshin:
`Extensions.MaybeAdjustFilePath` is a good one, but it works best when you pass root path to it (contextually sensible one, like project directory in most cases). Still, better than nothing.

PR #199, opened 2018-09-10
Minor slashes fix improvements for #193

andrew-boyarshin:


andrew-boyarshin:
@hvanbakel they should. I will do tests later this week on Linux.

PR #200, opened 2018-09-15
Split Project2015To2017 into Core and Migrate2017

andrew-boyarshin:
* Move almost everything to a new _Core_ project
* Move VS15-related transformations to _Migrate2017.Library_
* Make `ProjectConverter.Convert` an extension method (95% compatible, new optional/recommended `ILogger` parameter added)
* Retain umbrella-project _Project2015To2017_ to preserve NuGet package
* Move all common properties to `Directory.Build.props`
* Give Project2015To2017Tests a proper name with a dot
* Update external NuGet dependencies

P.S. don't worry about 191 changed files - that's mainly moved files.


Directory.Build.props
```diff
@@ -0,0 +1,17 @@
+<Project>
```
hvanbakel: Sweet! This was on my list for quite some time.

Project2015To2017.Core/Project2015To2017.Core.csproj
```diff
@@ -0,0 +1,23 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard1.3;netstandard2.0</TargetFrameworks>
+    <RootNamespace>Project2015To2017</RootNamespace>
```
hvanbakel: We don't need to do this now but I like having the namespaces line up with the project names.

andrew-boyarshin: Me too, but that's for maintaining (mostly) stable public API surface.

andrew-boyarshin:
Since no issues are raised, I'll rebase this against latest master and merge this later today.

PR #201, opened 2018-09-17
AppVeyor improvements

andrew-boyarshin:
* Rewrite AppVeyor configuration related to actual build & test
* Improve CI performance
* Use SourceLink
* Simplify project files
* Don't generate nupkg during local development

Note: I'd like to fix Project2015To2017Tests namespaces before 4.0 (API breaking changes) release.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -12,6 +12,15 @@
 		<PackageTags>dotnet csproj fsproj vbproj msbuild conversion vs2015 vs14 vs15 vs2017</PackageTags>
 		<Version>3.0.2</Version>
 		<AssemblyVersion>3.0.0.0</AssemblyVersion>
+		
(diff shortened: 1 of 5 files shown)


hvanbakel:
You can rename anything in the test project, we don’t package that up so you can’t break anyone. Same for the console applications.

andrew-boyarshin:
No, I meant doing that not in this PR but before 4.0.

andrew-boyarshin:
And there is one more thing to do before 4.0 - move VS15-specific diagnostics to Migrate2017 project.

hvanbakel:
`SourceLink` will allow debugging I assume? So VS can download pdbs when they’re missing?

andrew-boyarshin:
Not just PDBs, but entire and precise source code related to commit from which NuGet was built.

PR #203, opened 2018-09-19
Move logic to respective places

andrew-boyarshin:


Project2015To2017/Facility.cs
```diff
@@ -0,0 +1,211 @@
...
+using Project2015To2017.Transforms;
+using Project2015To2017.Writing;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.IO;
+using System.Linq;
+
+namespace Project2015To2017
+{
+	public class Facility
```
hvanbakel: This is moving it down into the shared nupkg so it can be reused there, right?

Project2015To2017/PatternProcessor.cs
```diff
@@ -0,0 +1,4 @@
+namespace Project2015To2017
+{
+	public delegate bool PatternProcessor(ProjectConverter converter, string pattern, ProcessSingleItemCallback callback, Facility self);
```
hvanbakel: Out of curiosity, any reason why you're not using a Func?

Project2015To2017/Facility.cs
andrew-boyarshin: Yes, that's right.

Project2015To2017/PatternProcessor.cs
andrew-boyarshin: They are mostly interchangeable, but delegate provides more explicit function signature _declaration_.

PR #213, opened 2018-11-13
Ignore 'Import'

baruchiro:


hvanbakel:
You cannot do it like this, `Import` has a meaning to bring in external targets too, those should be removed as they now belong in the generated targets in the intermediate output path.

Have a look at some of the test cases, you’ll see that there are other uses of `Import` too so this will have to be a smarter check.

PR #226, opened 2019-01-12
4.0 RC

andrew-boyarshin:
* Migration wizard (very advanced, try it out yourself!)
* Modernization
* Turn legacy Project2015To2017.Console into stub
* Refactorings under the hood
* Fix mistakes with handling assembly information files
* Make output less noisy by default

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -10,7 +10,7 @@
 		<Copyright>Copyright Hans van Bakel</Copyright>
 		<Description>Tool for converting a MSBuild project file to VS2017 format and beyond.</Description>
 		<PackageTags>dotnet csproj fsproj vbproj msbuild conversion vs2015 vs14 vs15 vs2017</PackageTags>
-		<Version>3.0.10</Version>
+		<Version>3.9.0</Version>
 		<AssemblyVersion>3.0.0.0</AssemblyVersion>
 		
 		<!-- SourceLink -->
--- a/Project2015To2017.Console/Program.cs
+++ b/Project2015To2017.Console/Program.cs
@@ -1,11 +1,5 @@
-using System.Collections.Generic;
-using System.Linq;
 using CommandLine;
 using Microsoft.Extensions.Logging;
-using Project2015To2017.Analysis;
-using Project2015To2017.Definition;
-using Project2015To2017.Migrate2017;
-using Project2015To2017.Transforms;
 
 namespace Project2015To2017.Console
 {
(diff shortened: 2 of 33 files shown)


andrew-boyarshin:
Whoops, turns out it is published (as 3.9.0) on NuGet. While I consider it to be stable (non-regressing at least), I think it would be better marked as prerelease.

hvanbakel:
I just unlisted them so we should be fine

andrew-boyarshin:
@mungojam the major part of this PR is Interactive Wizard, but that's only in **.Migrate2017.Tool**, for obvious reasons. The library is improved too, but not as much.

andrew-boyarshin:
I've discovered a few issues. One is related to this PR (missing logger parameter in constructor of the new transform). Others are related to Unix support (paths are still a major issue) in FileTransformation. I'll fix them in next PRs.
Also in my next PR I will remove netstandard1.3 target, since 2.0 is much better and the supported range is only a bit smaller (who even remembers .NET Core 1.x nowadays? Or has 4.5 but doesn't have at least 4.6.1?). Dependencies have bumped their requirements.

Project2015To2017.Console/Program.cs
```diff
@@ -19,67 +13,9 @@ static void Main(string[] args)
...
-				convertedProjects.AddRange(projects);
-			}
-
-			System.Console.Out.Flush();
-
-			foreach (var project in convertedProjects)
-			{
-				analyzer.Analyze(project);
-			}
+			logger.LogError("csproj-to-2017 is obsolete. Use Project2015To2017.Migrate2017.Tool (dotnet migrate-2017).");
```
hvanbakel: Can we include a step by step guide for users? 
So dotnet install...
Dotnet migrate-2017 migrate bla.csproj

That way if they encounter this they don’t need to open a browser but can just run what it says there.

Directory.Build.props
```diff
@@ -10,7 +10,7 @@
 		<Copyright>Copyright Hans van Bakel</Copyright>
 		<Description>Tool for converting a MSBuild project file to VS2017 format and beyond.</Description>
 		<PackageTags>dotnet csproj fsproj vbproj msbuild conversion vs2015 vs14 vs15 vs2017</PackageTags>
-		<Version>3.0.10</Version>
+		<Version>3.9.0</Version>
```
hvanbakel:
```suggestion
		<Version>4.0.0-beta01</Version>
```

andrew-boyarshin:
I'll merge this in a couple of days and publish new PR with changes mentioned above.

PR #235, opened 2019-03-11
Keep linked app.config

MO2k4:
Hi, 

first off thank you for this great tool. I stumpled upon a minor problem for my migrated projects.

I am using linked app.configs in my test scenarios, so that i can reuse configuration, but after migrating them they are gone and i have to reconfigure them.

Why are *.config-Files skipped? I found this issue #155 but i do not understand why this is a problem.


andrew-boyarshin:
Yes, it appears the decision was based on a bad assumption. Merged, will make it into 4.1.0.

PR #236, opened 2019-03-12
Added direct support for removing transitive package references

MO2k4:
Hey,

i've got another request. 

After migration it would be great to use the new transitive way for nuget package references. So it would be nice to remove unnecessary references.

I added another command line argument for migration to clean up the references.

What do you think about it? I've got already a console application to get the output for removable references [AnalyzeDotnetProjects](https://github.com/MO2k4/AnalyzeDotNetProject)

### DIFF
--- a/Project2015To2017.Core/CleanUp/Dependency.cs
+++ b/Project2015To2017.Core/CleanUp/Dependency.cs
@@ -1,0 +1,50 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Project2015To2017.CleanUp
+{
+	internal class Dependency : IEquatable<Dependency>
+	{
+		public Dependency(string name, string version)
+		{
+			Name = name;
+			Version = version;
+		}
+
+		public string Name { get; }
+
+		public string Version { get; }
+
+		public string Parent { get; set; }
+
+		public List<Dependency> Children { get; set; } = new List<Dependency>();
+
+		public List<Dependency> ContainingPackages { get; set; }
+
+		public bool Equals(Dependency other)
+		{
+			if (ReferenceEquals(null, other)) return false;
+			if (ReferenceEquals(this, other)) return true;
+			return string.Equals(Name, other.Name) && string.Equals(Version, other.Version);
+		}
+
+		public override bool Equals(object obj)
+		{
+			if (ReferenceEquals(null, obj)) return false;
+			if (ReferenceEquals(this, obj)) return true;
+			if (obj.GetType() != this.GetType()) return false;
+			return Equals((Dependency)obj);
+		}
+
+		public override int GetHashCode()
+		{
+			unchecked
+			{
+				return ((Name != null ? Name.GetHashCode() : 0) * 397) ^ (Version != null ? Version.GetHashCode() : 0);
+			}
+		}
+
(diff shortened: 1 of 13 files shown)


hvanbakel:
We have had discussions on this before and at the time decided that it would be better placed in a different tool. I can see people running this over and over again where a migration tool is often a one off. 

I think we can revisit this, the question is how long it takes on bigger projects as there’s a bunch of file scanning/parsing that needs to happen to make this work. I will say though that it still feels better suited to be a different tool. I’ll have a look at the code later today.

andrew-boyarshin:
As a first feedback: the idea is great, implementation seems advanced enough, but the major drawback is the dependency on MSBuild. This project strives to be small and pretty self-contained. That is, being able to run on a machine with only .NET Core runtime installed. We could've just used sth like Buildalyzer and gain absolute correctness of property resolution, instead, we imported only a tiny part of MSBuild code to resolve `Condition` expressions, and that's it. We don't require presence of .NET SDKs (massive dependency!). Thus, this PR is unlikely to be merged, unless rewritten to use already existing ProjectReference abstractions instead of MSBuild.
@hvanbakel what is your opinion on the matter?

andrew-boyarshin:
It appears mobile GitHub doesn't update page contents. :)

hvanbakel:
I'll have an in depth look at the code later but a lot needs to happen to
get this right. Good tools do one thing and do that thing properly, you
cannot and should not roll everything into one.

That being said, I don't think we've ever gotten a pr for the feature as
opposed to a 'wouldnt it be nice' so let's keep an open mind

mungojam:
I'd like for this feature to be there, but I'd be just as happy for it to be another nuget library that I could just add a custom transform for. It's going to be a useful tool to run from time to time on any code base or could be part of a CI build to highlight redundancies

I'd happily have a dependency on buildalizer as it seems solid but I'm now used to loads of dependencies as I work in python and node at the moment

mungojam:
Having said that I don't agree with the simple approach of stripping all dependencies if they are covered transitively because of [this potential risk]( https://github.com/NuGet/Home/issues/5715) which I raised a while back but nobody seems to get 😥

hvanbakel:
Having looked at the PR I don’t want to bring this responsibility in. The reasons for this are:

1. I don’t think this belongs here as it’s not a part of the migration per se, having this as a separate tool makes more sense
2. The amount of dependencies as mentioned by @andrew-boyarshin 
3. The fact that it runs command line tools makes it ‘brittle’ in my opinion
4. The fact that I’m fairly convinced that the logic itself is not 100% safe. What do you do when A is a transitive dependency but is installed at a different (lower/higher) version. That all becomes a bit arbitrary imho and I can see plenty of issues emerging as a result of that which I wouldn’t want to support (certainly not given how external to the tool this is).

I very much appreciate the fact that you made a PR out of it though. I would suggest wrapping this in a global tool of your own and pushing that out. They can complement each other nicely.

MO2k4:
Thanks for your feedback, i will take a look at the points you mentioned.

@hvanbakel 
4. I am only removing the dependency if Name (Id) and Version are equal, thats why i implemented IEquatable only for this two.

I already did a seperat tool [AnalyzeDotNetProject](https://github.com/MO2k4/AnalyzeDotNetProject) - i will do some cleanup and take your feedback in account. Maybe i get a different approach without using msbuild and the command line tools.

@mungojam thanks for your feedback and the link towards this problem.

PR #244, opened 2019-04-27
Apply standard formatting across solution

mungojam:
I hope this can be a step towards avoiding accidental formatting changes in future. Up to now there has been a legacy of inconsistent formatting which is a problem when then trying to format our own changes using ctrl+k+d or whatever other tool you like to use.

I've manually done two things:
- Run dotnet-format
- Use ctrl+k+d on each .csproj file

I'm not an expert at `editorconfig` files, nor am I hugely opinionated about which conventions we use (spaces or tabs etc., I don't really mind). But I think it would be great to get things standardised as part of a check-in hook or just a validation that dotnet-format doesn't caused any changed when run in the build script.

### DIFF
--- a/Project2015To2017.Console/Project2015To2017.Console.csproj
+++ b/Project2015To2017.Console/Project2015To2017.Console.csproj
@@ -1,18 +1,18 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
-	<PropertyGroup>
-		<TargetFramework>netcoreapp2.1</TargetFramework>
-		<PackAsTool>True</PackAsTool>
+  <PropertyGroup>
+	<TargetFramework>netcoreapp2.1</TargetFramework>
+	<PackAsTool>True</PackAsTool>
 
-		<PackageId>Project2015To2017.Cli</PackageId>
-		<Product>Project2015To2017.Cli</Product>
-		<OutputType>Exe</OutputType>
-		<ToolCommandName>csproj-to-2017</ToolCommandName>
-	</PropertyGroup>
+	<PackageId>Project2015To2017.Cli</PackageId>
+	<Product>Project2015To2017.Cli</Product>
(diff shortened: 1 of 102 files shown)


andrew-boyarshin:
I'll have a look at it tomorrow. 👍

andrew-boyarshin:
I've verified that no functional change is present. Would you mind if I also ran JetBrains Rider code cleanup and add to this PR? To make it final once and for all.

mungojam:
Go for it (I love resharper by the way). I think rider can pick up settings from editorconfig, so hopefully over time the two methods will align

hvanbakel:
I don’t see an `editorconfig` file, correct? 
If so, how do we want to keep the code in sync with this formatting? Different tools have different preferences :).

mungojam:
There is one, you added it 🙂

https://github.com/hvanbakel/CsprojToVs2017/blob/master/.editorconfig

mungojam:
> Different tools have different preferences

That's true, do you use resharper/rider? For settings beyond the current scope of editorconfig, we could add solution wide resharper preferences. At least @andrew-boyarshin and I are using it.

mungojam:
@andrew-boyarshin Would you mind if I also ran JetBrains Rider code cleanup

Actually, by code cleanup, do you mean just formatting or also things like removing unused methods? I wouldn't want that sort of API change here because previously it has broken my custom use of the library (when AssemblyName was removed). Unused usings removal would be fine.

hvanbakel:
> There is one, you added it 🙂
> 
> https://github.com/hvanbakel/CsprojToVs2017/blob/master/.editorconfig

😄  I saw the comment about editorconfig and figured there must not have been one given the issues addressed here but I guess not every tool used leverages it.

mungojam:
I'm gonna merge this formatting PR, @andrew-boyarshin, you can always add rider formatting later if you like. I read that the new resharper can export it's settings to editorconfig so that might be useful in future

Issue #252, opened 2019-06-26
Migration removes all .cs files


josh-degraw:
I've used this tool before without issue, but I just ran it on a solution with several projects and encountered a weird issue where it added a line in the modified `.csproj` files that for each `MyFile.cs` it found, it added the line:

```xml
<Compile Remove="MyFile.cs"/>
```

which, as the syntax implies, completely removed the files from the project-- they were all empty. It was simple enough to fix, but it seems like something that just shouldn't happen in the first place. This _seems_ like a similar issue to #210, but in my case these were all just class library projects utilized in an `ASP.NET` `MVC` project, not WinForms, and none of the files were `SubType`'d.

mungojam:
Were the files included in the original project? It's meant to do this just for files that weren't included in the original so that they don't automatically get included in the converted project

superstrom:
I've seen this happen if you run migrate on an already VS2017-style project.

josh-degraw:
@mungojam No they weren't explicitly referenced in the project, and @superstrom I think that might have been the case. Is there any sort of fix that can be done for that? I had manually updated a few of the projects, and I was hoping to clean up some elements that were still left over from that in the process.

mungojam:
What command did you run to do the conversion and did you use the wizard? I thought it blocked conversions for VS2017 projects and made you run modernise instead

If you're able to share an example that would be really helpful

@andrew-boyarshin

superstrom:
I can reproduce with
```
dotnet new console
dotnet migrate-2017 wizard
```

josh-degraw:
@superstrom Yep, that's basically how it happened for me, and the output looks just like that.

hvanbakel:
Yeah I think I can see how this happens:
migrate looks for all the files in the csproj
there are none, so all the files on disk must be excluded from the project thus insert all those removes 😄 

I guess we can fix that, yes.

hvanbakel:
Would you mind trying with:
https://www.nuget.org/packages/Project2015To2017.Migrate2017.Tool/4.1.0-beta.3

It does seem to work for me given the repro case above.

superstrom:
Yeah, the update is working for me.

PR #253, opened 2019-06-27
#252 fix

hvanbakel:


hvanbakel:
and I just learned that PRs also get published so the nuget package is already out there... 

Filing a separate bug for that.

Project2015To2017.Migrate2017.Library/Transforms/FileTransformation.cs
```diff
@@ -47,6 +47,11 @@ public FileTransformation(ILogger logger = null)
 
 		public void Transform(Project definition)
 		{
+			if (definition.IsModernProject)
```
andrew-boyarshin: It's better to add `ILegacyOnlyProjectTransformation` interface to `FileTransformation` instead of putting an `if` statement here. This way, user can override this by command-line switch.

hvanbakel:
That's a really good point, wasn't around a PC so this was done through the
web UI :smile: I'll fix that later

Project2015To2017.Migrate2017.Library/Transforms/FileTransformation.cs
andrew-boyarshin: It's already `ILegacyOnlyProjectTransformation`, so the problem is actually more profound. I'm taking over the issue to investigate.

hvanbakel:
This is a good reminder I forgot about this one. Do you want me to look at
it first?

andrew-boyarshin:
@hvanbakel no, see my latest comment to the code line.

andrew-boyarshin:
Superseded by my latest commit on master branch with a better fix:
1. Incorrect transformation set used in wizard logic
2. No guard in `FileTransformation` against default wildcards in modern projects

PR #255, opened 2019-07-19
Create migrate-2019 for 4.1-RC

### LINKED ISSUE #254: Creating a PR also publishes nuget packages if it contains a revved version

andrew-boyarshin:
* Refactor `migrate-2017` to reduce code duplication
* Split large files into multiple small ones
* Move `Service` item filtering to separate transformation
* `Service` is an item, not a property (bugfix)
* Remove duplicates from input paths (e.g. when globbing like `**\*` a single project could be added twice: from actual glob search and from solution traverse)
* Update MSBuild.Sdk.Extras versions for both .NET Core 2.x & 3.x
* Update AppVeyor configuration (fix #254)

@hvanbakel one more NuGet package is required for the new `dotnet migrate-2019` global tool. We can't reuse legacy `.Console` API key for obvious reasons.


andrew-boyarshin:
@mungojam related to your comment <https://github.com/hvanbakel/CsprojToVs2017/pull/243#issuecomment-487751092>:
That's strange. My response on that thread was incorrect. You don't need to set target execution moment on your pre/post-transforms passed through `ConversionOptions`, they are executed separately from topologically sorted ones. I believe the reason you saw what you saw on your screenshot is that you've added the same transforms to TransformationSet. It's not necessary. Moreover, this makes your custom pre/post transforms run twice (as part of transformation set and separately from `ConversionOptions`).
I'd deprecate `PreDefaultTransforms` and `PostDefaultTransforms` at all, and force everything to be part of one transformation plan graph (that is, via `ITransformationSet`). This would remove such ambiguities.

mungojam:
@andrew-boyarshin, by that point I had given up on trying to use the conversion options route as the transforms would get run twice due to the way the Facility class processed things (sorry can't remember exact reason at the moment).

I think the concept of pre and post transforms covers a lot of scenarios, but as long as the way I have put in the readme still works then I'm ok with dropping the ConversionOptions pre/post in order to simplify the logic.

hvanbakel:
I’ll have a look at it over the weekend, I just tot an email about an expiring NuGet api key anyway so I’ll create the 2019 one too.

The downside from creating a completely new tool though is that users will need a new package to install. And since the problem is not remotely as big as it was for 2015 to 2017 I don’t expect a lot of people to install that. Should we maybe check if one package can contain multiple commands?

andrew-boyarshin:
@hvanbakel as long as we recommend using `migrate-2019` instead of `migrate-2017` (e.g. in README.md) no further action will be needed. `migrate-2019` is as functional as `migrate-2017`, transformation sets differ only by a single WPF/WinForms transformation, 99% of the code is shared and reused. It is possible to migrate from 2010 to 2019 in a single `migrate-2019` command.

andrew-boyarshin:
OK then, pending new NuGet package for 4.1 release. Meanwhile, I'll update the README and add some `[Obsolete]` annotations.

Project2015To2017.Core/Definition/Project.cs
```diff
@@ -88,5 +88,35 @@ public DirectoryInfo NuGetPackagesPath
...
+				if (ReferenceEquals(x, y)) return true;
+				if (x is null) return false;
+				if (y is null) return false;
+				if (x.GetType() != y.GetType()) return false;
+				return string.Equals(x.ProjectName, y.ProjectName, StringComparison.InvariantCultureIgnoreCase) &&
+				       string.Equals(x.FilePath.FullName, y.FilePath.FullName,
+					       StringComparison.InvariantCultureIgnoreCase);
+			}
+
+			public int GetHashCode(Project obj)
```
hvanbakel: Why not reuse the `FilePathEqualityComparer` defined below?

Project2015To2017.MigrateXXXX.Tool/CommandLogic.cs
```diff
@@ -0,0 +1,131 @@
...
+
+		public void ExecuteAnalyze(
+			IReadOnlyCollection<string> items,
+			ConversionOptions conversionOptions,
+			AnalysisOptions analysisOptions)
+		{
+			facility.ExecuteAnalyze(items, conversionOptions, analysisOptions);
+		}
+
+		public void ExecuteWizard(
```
hvanbakel: This signature should be reordered to match the other ones. Make `sets` the last param in stead of the first like the other ones.

Project2015To2017.Migrate2019.Library/Vs16TransformationSet.cs
```diff
@@ -32,7 +33,7 @@ private Vs16TransformationSet()
 					conversionOptions.TargetFrameworks,
 					conversionOptions.AppendTargetFrameworkToOutputPath),
 				// VS15 migration
```
hvanbakel: should be VS16 here

Project2015To2017.MigrateXXXX.Tool/CommandLogic.Wizard.UnknownTargetFramework.cs
```diff
@@ -0,0 +1,90 @@
...
+						}
+
+						return true;
+					}
+
+					break;
+				}
+			}
+
+			Log.Information("Please, enter target frameworks to use (comma or space separated):");
```
hvanbakel: I would include a full fx, net standard and net core example here

hvanbakel:
New keys are here:
https://github.com/hvanbakel/CsprojToVs2017/commit/ff1c66c179c526120b4b046aa227d33e18fe8e8f

The 2019 key is in comment to be uncommented here.

Project2015To2017.Core/Definition/Project.cs
andrew-boyarshin: Because `FilePathEqualityComparer` is an `IEqualityComparer<Solution>`, and this takes care of `Project`. Moreover, in this `IEqualityComparer<...>` I've opted into comparing `ProjectName` too.
I could, I guess, factor out `IEqualityComparer<FileInfo>` and reuse it in both comparers, but I'm not sure it is worth the effort.

hvanbakel: Agreed, it was for comparing the paths yes but no big deal indeed

PR #257, opened 2019-07-31
In cases where AssemblyInfo is being preserved and there are multiple…

jlegan:
… AssemblyInfo files present (very common in GitVersion implementations where there is a single common AssemblyInfo file containing version info and an AssemblyInfo file containing library specific details), AssemblyDetails will be null (multiple files not currently supported) but there is no null check in NuGetPackageTransformation. Added null propagation on the assemblyAttributes object. If time allows I will add support for multiple assemblyinfo files but at a minimum this fixes the null reference exception when assembly info details are being preserved.

### DIFF
--- a/Project2015To2017.Core/Transforms/NuGetPackageTransformation.cs
+++ b/Project2015To2017.Core/Transforms/NuGetPackageTransformation.cs
@@ -57,10 +57,10 @@ namespace Project2015To2017.Transforms
 			{
 				//Id does not need to be specified in new project format if it is just the same as the assembly name
 				Id = rawPackageConfig.Id == "$id$" ? null : rawPackageConfig.Id,
-				Version = PopulatePlaceHolder("version", rawPackageConfig.Version, assemblyAttributes.InformationalVersion ?? assemblyAttributes.Version),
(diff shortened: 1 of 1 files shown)


andrew-boyarshin:
@jlegan regarding your wish to improve support for shared (common, global) assemblyinfo files. They are already supported (I regularly test the project on [ConfuserEx](https://github.com/yck1509/ConfuserEx) and [Dopamine](https://github.com/digimezzo/Dopamine), both have shared assemblyinfo files). The degree of that support however is minimal: preserve all assemblyinfo files, do not read information from them. I introduced such behavior back in #139, to not to introduce errors during conversion. There aren't much things that can be done to improve the situation (or they are quite complicated).

- Read all assemblyinfo files and transform read properties into includable MSBuild project with a single `PropertyGroup`.
- Read shared assemblyinfo files and add specific properties (e.g. `GenerateAssemblyCompanyAttribute`) for each shared property

Note, that these require making a lot of assumptions about the structure of the project. E.g. in ConfuserEx the shared assemblyinfo file might not exist at the time of conversion (it is being generated during build process by one of the projects being built).

Thanks you for your contribution, I can't believe such bug existed and wasn't noticed before.

PR #258, opened 2019-08-06
Transform assembly signing attributes

LTruijens:
Siging of an assembly can also be done using assembly attributes in the assemblyinfo.cs

[assembly: AssemblyDelaySign(true)]
[assembly: AssemblyKeyFile(@"..\..\PublicKey.snk")]

With this pull request these attributes get also transformed

### DIFF
--- a/Project2015To2017.Core/Definition/AssemblyAttributes.cs
+++ b/Project2015To2017.Core/Definition/AssemblyAttributes.cs
@@ -91,6 +91,21 @@ namespace Project2015To2017.Definition
 			(this.FileVersion != null && this.FileVersion.Contains("*")) ||
 			(this.Version != null && this.Version.Contains("*"));
 
+		public bool IsSigned => 
+			this.KeyFile != null;
+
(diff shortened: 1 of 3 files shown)


Project2015To2017.Core/Definition/AssemblyAttributes.cs
```diff
@@ -121,7 +136,7 @@ public string GetAttribute(Type attributeType)
 
 			//Make the assumption that it just has a single string argument
 			//because all of the attributes we currently look for do have
-			return att?.ArgumentList.Arguments.Single().ToString().Trim('"');
+			return att?.ArgumentList.Arguments.Single().ToString().TrimStart('@').Trim('"');
```
mungojam: I don't believe this change will achieve anything. The `@` at the start e.g. `@"c:\something"` is just for reading the typed string, the resultant string in memory will be the same as if somebody had escaped slashes directly `"c:\\something"`.

PR #288, opened 2021-10-16
Update W032OldLanguageVersionDiagnostic.cs

Jothay:
Updated language versions check


Jothay:
Per the note in the file regarding Latest + 2 versions back, I was just following the note

PR #290, opened 2022-03-23
Target net60

dustinsoftware:


hvanbakel:
@dustinsoftware thank you for the pr, but what does this get you since it already targets net standard? Just curious

hvanbakel:
Are you looking for the nupkgs in particular? I probably need to update keys if you are.

dustinsoftware:
Yeah dotnet tool install will pull from nuget so those need an update..

hvanbakel:
I'll see if I can get those back publishing over the weekend

